PR #5, opened 2021-05-27
add classes for Author, Footer, Provider and Image and add useful methods

Vendicated:
additionally returns self in all setters so you can chain

I didn't test this because the installer gets stuck on Installing indefinitely when enabling use dex from storage so no clue whether it works correctly lmao

I think it'd be good if there was a `change git upstream` option in installer so you can use your fork in it and didnt have to use dex from storage. I'll try to add it but i have no idea how flutter works so idk if i'll figure it out lmao

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/entities/MessageEmbed.java
+++ b/Aliucord/src/main/java/com/aliucord/entities/MessageEmbed.java
@@ -8,6 +8,7 @@ package com.aliucord.entities;
 import com.aliucord.Main;
 import com.discord.models.domain.ModelMessageEmbed;
 
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


Juby210 (approved):
Well changes are okay but it'll break existing plugins that use embeds so I'll merge it later.

Vendicated:
will it really break existing plugins? all that changed is the return type and since it used to return void its not like anyone relied on the return type.

or does java just shit itself?

either way wouldn't it be better to merge now than later because later there'll be more plugins that break

oh and lastly could you fix the use local dex issue so I can test my changes properly? I wanna make sure it's all working correctly

PR #7, opened 2021-05-31
fix borked plugin cards in amoled

Swishilicous:
i also may have changed the search input a little bit because i think the old version looked a bit off but you can change it back if you want


WinstonSepruko:
> fix borked cards in amoled

omg thanks finally

MulverineX:
Wonderful, I'm guessing this fixes them for Juby's LSPosed Themer too right?

Swishilicous:
I would have no idea but I assume his themer can hook into the fields I referenced in the commit.

MulverineX:
Oh it still doesn't then, dang

Juby210:
This PR is still not merged because it fixes plugin cards in amoled by breaking them on all other themes.

PR #9, opened 2021-05-31
Add HttpUtils Request/Response classes to ease making Http Requests

Vendicated:
Adds RequestBuilder kinda similar to the Powercord one to ease sending HttpRequests. Does not have querystring support yet and I also haven't tested it yet so its still a draft

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/Http.java
+++ b/Aliucord/src/main/java/com/aliucord/Http.java
@@ -11,7 +11,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 
-public class HttpUtils {
+public class Http {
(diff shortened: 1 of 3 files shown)


Swishilicous:
You should also make methods with the auth token as the authorization header like powercord so calling direct http requests to discord is easier.

Juby210 (commented):
Is this ready to merge or still `draft`?

Vendicated:
i think it has everything thats needed now, just have to test to make sure it's all working alright

Vendicated:
Ok ready for merge now, tried Json GET and POST and both worked well. I changed the updater to using the new methods

This should not break plugins as i left the methods in and marked them as deprecated, or should we just remove them?

Juby210 (approved):

PR #10, opened 2021-06-02
Clyde Customizations and other improvements

X1nto:
# Summary
Plugin developers can now customize Clyde's name and avatar, making it more user friendly and unique per plugin.

## Usage
`CommandsAPI.CommandResult` got 2 new parameters in the constructor, those are `username` and `avatarUrl`. `username` is used for the bot name and `avatarUrl` is the URL used for bot's avatar. Note: URL set in `avatarUrl` must point to raw image and not a redirect.

## Example
```java
MessageEmbed embed = new MessageEmbedBuilder()
        .setTitle("Drac")
        .setDescription("Haha Jonathan, you are banging my daughter.")
        .build();

return new CommandsAPI.CommandResult(null, embed, "Drac", "https://i.kym-cdn.com/entries/icons/original/000/037/312/cover2.jpg", false)
```

# Breaking Changes
MessageEmbed has been renamed to MessageEmbedBuilder and the `embed` field is now private, instead plugin developers should use the `build()` method which returns the Discord's `MessageEmbed`.

# Other Info
Also updated build script including `targetSdk`, `compileSdk` and `buildToolVersion`. Installer still targets sdk 29 because of scoped storage enforcements when targeting API 30.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/api/CommandsAPI.java
+++ b/Aliucord/src/main/java/com/aliucord/api/CommandsAPI.java
@@ -8,7 +8,6 @@ package com.aliucord.api;
 import android.os.Handler;
 import android.os.Looper;
 
-import com.aliucord.CollectionUtils;
 import com.aliucord.Main;
 import com.aliucord.Utils;
 import com.aliucord.utils.ReflectUtils;
@@ -49,6 +48,8 @@ public class CommandsAPI {
         public String content;
         public List<MessageEmbed> embeds;
         public boolean send;
+        public String name;
+        public String avatarUrl;
 
         public CommandResult() { this(null); }
         public CommandResult(String content) {
@@ -57,6 +58,19 @@ public class CommandsAPI {
         public CommandResult(String content, List<MessageEmbed> embeds, boolean send) {
             this.content = content;
             this.embeds = embeds;
+            this.send = send;
+        }
+        public CommandResult(String content, List<MessageEmbed> embeds, String name, boolean send) {
+            this.content = content;
+            this.embeds = embeds;
+            this.name = name;
+            this.send = send;
+        }
+        public CommandResult(String content, List<MessageEmbed> embeds, String name, String avatarUrl, boolean send) {
+            this.content = content;
+            this.embeds = embeds;
+            this.name = name;
+            this.avatarUrl = avatarUrl;
             this.send = send;
         }
     }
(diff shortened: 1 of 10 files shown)


Vendicated:
You should probably split these changes into seperate prs so that the commit messages make more sense

Aliucord/src/main/java/com/aliucord/api/CommandsAPI.java
```diff
@@ -59,6 +60,19 @@ public CommandResult(String content, List<MessageEmbed> embeds, boolean send) {
             this.embeds = embeds;
             this.send = send;
         }
+        public CommandResult(String content, List<MessageEmbed> embeds, String name, boolean send) {
+            this.content = content;
+            this.embeds = embeds;
+            this.name = name;
+            this.send = send;
+        }
+        public CommandResult(String content, List<MessageEmbed> embeds, String name, String avatarUrl, boolean send) {
```
Vendicated, line 69: send should be the third argument tbh, you will more often want to specifiy send than a custom profile

X1nto:
> You should probably split these changes into seperate prs so that the commit messages make more sense

Hmm, yeah it might be a good idea to do so.

Vendicated (changes requested):

Vendicated:
i guess squash commit 1 and 4 and then juby can just merge these 3 commits like they are

Aliucord/src/main/java/com/aliucord/api/CommandsAPI.java
X1nto: Yeah but we have another overload for the constructor which doesn't accept any customizations for clyde https://github.com/Aliucord/Aliucord/blob/39ebdb38d34aeb837a55b299d2393b97cc6a0361/Aliucord/src/main/java/com/aliucord/api/CommandsAPI.java#L58-L62  which is exactly for the case you're talking about

Vendicated: but then it doesnt make sense that argument position changes in the overload, so i still think send should be the third param even in the overload with name and avatarUrl, and also maybe change name -> username

Vendicated, line 69:
```suggestion
        public CommandResult(String content, List<MessageEmbed> embeds, boolean send, String username, String avatarUrl) {
```

X1nto: Aight, gotcha.

Vendicated:
very epic

Vendicated (approved):

Juby210 (approved):

PR #15, opened 2021-06-04
Make ui more responsive via toasts/state text changes

Vendicated:
or the title i prefer: Toastify me!

There are also Logger.error(context, message, error) and Logger.info(context, message) now that will automatically display a toast

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/widgets/UpdaterPluginCard.java
+++ b/Aliucord/src/main/java/com/aliucord/widgets/UpdaterPluginCard.java
@@ -7,8 +7,6 @@ package com.aliucord.widgets;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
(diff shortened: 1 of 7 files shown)


Aliucord/src/main/java/com/aliucord/widgets/UpdaterPluginCard.java
```diff
@@ -43,6 +39,7 @@ public UpdaterPluginCard(Context context, String plugin, Runnable forceUpdate) {
         setContentPadding(padding, padding, padding, padding);
 
         Plugin p = PluginManager.plugins.get(plugin);
+        assert p != null;
```
Vendicated, line 42: maybe this should be changed to a more appropriate error? but it should never happen anyway so whatever

Juby210 (approved):

PR #22, opened 2021-06-08
Improved notification API

X1nto:
# Summary
Replaced direct field calls with getters and setters as they're more convenient, also added documentation. Some of docs are most likely invalid so I added a `FIXME` annotation to them.

# Notes
- This **will** break the current implementation

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/entities/NotificationData.java
+++ b/Aliucord/src/main/java/com/aliucord/entities/NotificationData.java
@@ -8,6 +8,10 @@ package com.aliucord.entities;
 import android.graphics.drawable.Drawable;
 import android.view.View;
 
+import androidx.annotation.IdRes;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import com.discord.app.AppComponent;
 import com.discord.models.sticker.dto.ModelSticker;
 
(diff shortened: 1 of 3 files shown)


X1nto:
> I prefer direct access as we don't really need to do anything in set/get methods here tbh.

Setters allow chaining though, which kind of gives a better syntax.

Vendicated (approved):

Aliucord/src/main/java/com/aliucord/entities/NotificationData.java
```diff
@@ -17,19 +21,279 @@
...
+     * Sets the auto dismiss period.
+     * @param autoDismissPeriodSecs Dismiss period in seconds.
+     * @return {@link NotificationData} for chaining.
+     */
+    public NotificationData setAutoDismissPeriodSecs(@Nullable Integer autoDismissPeriodSecs) {
+        this.autoDismissPeriodSecs = autoDismissPeriodSecs;
+        return this;
+    }
+
+    //FIXME Documentation is invalid.
```
Juby210, line 244:
```suggestion
```
I don't see anything to fix here.

Aliucord/src/main/java/com/aliucord/entities/NotificationData.java
```diff
@@ -17,19 +21,279 @@
...
+    /**
+     * @return Valid screens
+     * @see NotificationData#setValidScreens(List)
+     */
+    public List<? extends c<? extends AppComponent>> getValidScreens() {
+        return validScreens;
+    }
+
+    //FIXME Documentation is invalid.
```
Juby210, line 253:
```suggestion
```
here too

Juby210 (changes requested):

Juby210 (approved):

PR #29, opened 2021-06-10
Add wrappers for obfuscated Discord classes

Vendicated:
i figured still using stubs is easier and more performant than using reflection, so instead we can just mark all stubs as deprecated

Is this good or would you change anything?


Vendicated:
well linking to the wrappers and vice versa in javadoc wont work because of the circular dependency introduced by adding Aliucord to the stubs project :/

PR #42, opened 2021-06-26
Actionbar buttons for Plugins + Crashes

WinstonSepruko:
This moves the "Clear logs" button to an action button in the toolbar on the crashes page + adds a button to open folders in both Plugins and Crashes. Plugins page button isn't the right colour yet, the Crashes ones appear to be (for me at least). Need to re-add the Crashes hint because it's currently commented out to prevent crashing.

Obviously not ready for merge, need to get button highlight working too.


Aliucord/src/main/java/com/aliucord/settings/Crashes.java
```diff
@@ -68,10 +74,60 @@ public void onViewBound(View view) {
         setActionBarTitle("Crash Logs");
 
         Context context = requireContext();
-        LinearLayout layout = (LinearLayout) ((NestedScrollView) ((CoordinatorLayout) view).getChildAt(1)).getChildAt(0);
-
         int p = Utils.getDefaultPadding() / 2;
-        layout.setPadding(p, p, p, p);
+        File folder = new File(Constants.BASE_PATH, "crashlogs");
+        File[] files = folder.listFiles();
```
Vendicated, line 79: you should move this into the clear button onClick handler and use folder.exists() instead to determine whether the folder exists.

Also I'm pretty sure calling button.setClickable(true) does nothing since that's the default, so you should either use button.setClickable(folder.exists()) or just don't add the button in the first place

WinstonSepruko:
I messed up the commits in my repo, closing because I need to fix stuff big time!

PR #46, opened 2021-07-03
Couple of README fixes and add two issue templates

NatSepruko:
Also adds a link to the support server when creating an issue!


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,48 @@
+name: Bug Report
+description: Create a bug report for Aliucord itself, to report bugs in plugins please go to our support server or the plugin's repository and make an issue there instead.
```
dweee, line 2: I'd say split the complex sentence into two separate ones since it is a bit long winded.

Juby210 (approved):

PR #60, opened 2021-07-25
Errors Not Showing Multiple Authors

Wolfkid200444:
Currently when a error is made by a plugin it will show the author but if a plugin has multiple authors it will only show the first author and not the other authors


Wolfkid200444:
Did the changes Veny asked for

PR #61, opened 2021-07-26
Support .aliu plugin extension

X1nto:


zt64 (approved):

Aliucord/src/main/java/com/aliucord/PluginManager.java
```diff
@@ -31,7 +31,7 @@
      */
     @SuppressWarnings({"JavaReflectionMemberAccess", "unchecked"})
     public static void loadPlugin(Context context, File file) {
-        String name = file.getName().replace(".zip", "");
+        String name = file.getName().replace(".zip", "").replace(".aliu", "");
```
nearlySplat, line 34: Couldn't this use the regexp OR?

```suggestion
        String name = file.getName().replace("\\.(?:zip|aliu)", "");
```

Vendicated:
since this isn't really an issue anymore now that you don't manually install plugins anymore and with the upcoming store and how Aliucord now warns you if it finds extracted plugins i dont think this is necessary anymore

PR #83, opened 2021-08-23
Added some kotlin extensions

X1nto:
Because java APIs are too limited.


Juby210:
They are not needed as Aliucord will receive them in API rewrite.

PR #102, opened 2021-09-15
Add sha1 Integrity check to Http#saveToFile & Plugin Updater

Vendicated:
- make saveToFile save to temp file and then move to the actual path. This ensures files aren't overwritten with garbage/nothing if something goes wrong
- Add option to verify integrity using md5sum 

The gradle plugin would then calculate and add m5sums to the updater.json

Draft because I still need to add it to the gradle plugin and also test that it works correctly


Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
```diff
@@ -29,6 +29,7 @@
         public String build;
         public String changelog;
         public String changelogMedia;
+        public String md5sum;
```
js6pak, line 32:
```suggestion
        public String sha1;
```

js6pak (changes requested):

PR #103, opened 2021-09-16
Add debuggable to app info

TymanWasTaken:
May or may not work I did this on mobile


hairyfred:
Should be added to the aliucord commit version in discord settings as EG XXXXXX_DEBUG

TymanWasTaken:
Screenshot_20210915-185842
It works, tell me if I should change anything

PR #106, opened 2021-09-18
Add links to the manifest

js6pak:
Adds a new field to the manifest.json, which looks like
```json
"links": {
	"github": "https://github.com/js6pak/WhoReacted"
}
```

It can have any key that you want, which can later be used by aliucord or other plugins.
Aliucord currently only uses `source` (link to a code source, like github or gitlab) and `github` which is a superset of `source`, which behaves the same way but has github icon (which is not implemented because there is no git icon in lytefast lol). They are used in plugin list and changelog footer.
Links could also be used for twitter or whatever

Now that I'm writing this, I don't think that `links` is a good name, maybe something like `metadata` would be more fitting?

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/PluginManager.java
+++ b/Aliucord/src/main/java/com/aliucord/PluginManager.java
@@ -11,6 +11,8 @@ import android.content.res.AssetManager;
 import android.content.res.Resources;
 
 import com.aliucord.entities.Plugin;
+import com.aliucord.patcher.Patcher;
+import com.aliucord.patcher.PineInsteadFn;
 import com.aliucord.utils.*;
 
 import java.io.*;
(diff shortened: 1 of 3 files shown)


Juby210 (approved):

Aliucord/src/main/java/com/aliucord/PluginManager.java
```diff
@@ -71,6 +73,8 @@ public static void loadPlugin(Context context, File file) {
 
                 manifest.name = pluginInstance.name;
                 pluginInstance.initialize(manifest);
+
+                Patcher.addPatch(pluginClass.getDeclaredMethod("getManifest"), new PineInsteadFn(callFrame -> manifest));
```
Vendicated, line 77: remove, make up for this inside adapter.onRepoClick instead

Vendicated (changes requested):

PR #111, opened 2021-10-05
Update to v96.0

Vendicated:
Breaks most plugins, so should not be merged until plugin devs updated their plugins

Changes:
- Remove all deprecated apis (old patcher, deprecated utils, etc)
- Utils.appContext is now Utils.getAppContext()
- Discord added more options to CommandOptions so all commands are broken


omxpro:
Yessir

PR #116, opened 2021-10-18
Improve TextInput

wingio:
Adds a smaller Text Input based on the new nickname textbox
image


Vendicated:
1) I don't like the name of this view it feels weird
2) Please add @NonNull / @Nullable annotations for parameters and return types, makes it easier to use in kotlin

PR #123, opened 2021-10-21
Improve plugins page

ItzOnlyAnimal:
- "find plugins" button that opens the Aliucord invite splash when no plugins are installed


ItzOnlyAnimal:
Okay it should be fine now

ItzOnlyAnimal:
gonna make new pr with multi select

PR #127, opened 2021-10-25
Only show enabled plugins by default in plugins command

zt64:


Vendicated:
don't think this is a good idea tbh. maybe a better idea would be making the output something like this

```
Enabled Plugins (42):
...

Disabled Plugins (2):
...
```

zt64:
I guess, I'll change it when I have time

PR #129, opened 2021-10-30
feat(CommandContext): role/channel option getters

DiamondMiner88:
Add missing command option getters


Aliucord/src/main/java/com/aliucord/entities/CommandContext.java
```diff
@@ -373,4 +385,66 @@ public User getUserOrDefault(String key, User defaultValue) {
         User val = getUser(key);
         return val != null ? val : defaultValue;
     }
+
+    /**
+     * Gets the Channel argument with the specified key
+     * @param key Key of the argument
+     */
+    @Nullable
+    public ChannelWrapper getCurrentChannel(String key) {
```
Vendicated, line 394: shouldn't this be getChannel not getCurrentChannel

PR #131, opened 2021-11-06
fix: enable settings btn of disabled plugins

DiamondMiner88:
Don't disable settings buttons on disabled plugin cards

As discussed here [here](https://discord.com/channels/811255666990907402/811261478875299840/904135001367531561) (Aliucord discord)

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/settings/Plugins.java
+++ b/Aliucord/src/main/java/com/aliucord/settings/Plugins.java
@@ -76,7 +76,7 @@ public class Plugins extends SettingsPage {
             }
 
             public void onToggleClick(boolean checked) {
-                adapter.onToggleClick(this, checked, getAdapterPosition());
+                adapter.onToggleClick(getAdapterPosition());
             }
 
             public void onUninstallClick(View view) {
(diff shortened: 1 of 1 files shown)


Vendicated:
I don't know if this is a good idea, since there are multiple plugins whose settings rely on being started  or restart themselves after you change settings (see for instance https://github.com/zt64/aliucord-plugins/blob/65440c8f947f5ba034d67a775c6e8759d191bc86/NoBurnIn/src/main/java/tk/zt64/plugins/noburnin/PluginSettings.kt#L48)

DiamondMiner88:
Maybe have it as an option in the @Aliucord annotation?

Vendicated:
when would you even want to change a plugins settings without having it enabled?

Vendicated:
As mentioned above this isn't a good idea because it will lead to unexpected behaviour. Making it an annotation param seems weird too because users will be very confused why some plugins have their settings enabled and some don't. Maybe it would make more sense to make a raw settings editor (probably could be a plugin) where it just makes a page with a text field per setting of a plugin and you can change these to whatever you want?

PR #133, opened 2021-11-06
Add logger field to Plugin class

zt64:


Vendicated:
tbh idk about this because from my experience non static logger on plugin class is meh. In most of my plugins I either have a top level logger with kotlin or make it a static field on my plugin class so I can use it from anywhere

PR #135, opened 2021-11-07
Migrate views to kotlin

ItzOnlyAnimal:
- Changes Button, DangerButton, Divider, SaveButton, TextInput, and ToolbarButton from java to kotlin


Vendicated:
1) Fix null types, context cannot be null
2) Use kotlin getters and setters more so that you can use direct access in kotlin
3) why

ItzOnlyAnimal:
> why
honestly I have no idea but kotlin
will fix when I have too much free time once again

ItzOnlyAnimal:
see #161

PR #138, opened 2021-11-08
Added R wrapper

mantikafasi:


Vendicated:
seeing as this doesn't work in kotlin and will break anyway if there's an update I don't think this makes much sense

PR #141, opened 2021-11-11
Add DiscordRequest to Http

wingio:
Automatically adds headers and allows you to remove the `https://discord.com/api/v9` base url


Vendicated:
can u make this a method in request instead of a seperate class

something like Req.addDiscordAuthHeaders()

PR #144, opened 2021-11-13
feat: `AliucordPlugin#requiresRestart`

DiamondMiner88:
`@AliucordPlugin` annotation property: prompt user to restart AC after
- Disabling manually
- Enabling manually
- Updating
- Uninstalling


Aliucord/src/main/java/com/aliucord/PluginManager.java
```diff
@@ -202,4 +204,17 @@ public static boolean isPluginEnabled(String name) {
     public static boolean isPluginEnabled(Plugin plugin) {
         return isPluginEnabled(MapUtils.getMapKey(plugins, plugin));
     }
+
+    /**
+     * Prompts the user to restart Aliucord because of a plugin that requires a restart
+     */
+    public static void promptRestart() {
```
Vendicated, line 211: Make this part of Utils perhaps? Maybe even with a custom message that defaults to the one you specified here? Also can't you just use Utils.appActivity as view

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -234,6 +234,8 @@ public void onToggleClick(ViewHolder holder, boolean state, int position) {
             String name = data.get(position).getName();
             PluginManager.togglePlugin(name);
             holder.card.settingsButton.setEnabled(state);
+            var p = Objects.requireNonNull(PluginManager.plugins.get(name));
```
Vendicated, line 237: data.get(position) = plugin, so just change the first line in this method instead of getting it like this

Vendicated (changes requested):

Vendicated:
Also, would be best to integrate it into PluginDownloader too

Vendicated:
Please fix conflict

Vendicated (approved):

PR #148, opened 2021-11-14
Port to 101.7

TymanWasTaken:
Literally nothing changed but ven said to so here we are

All plugins seem to work fine (at least mine), the only errors I can see were there in 101.3 anyway


Juby210:
android doesn't allow you to downgrade packages, beta version code is always lower than alpha on same major version, pointless PR

Vendicated:
superseded by #156

PR #151, opened 2021-11-16
Add missing param in doc

scrazzz:


Aliucord/src/main/java/com/aliucord/Http.java
```diff
@@ -252,10 +252,11 @@ public static Request newDiscordRequest(String route) throws IOException {
         /**
          * Performs a request to a Discord route
          * @param route A Discord route, such as `/users/@me`
+         * @param method <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods">HTTP method</a>
          * @throws IOException If an I/O exception occurs
          */
         public static Request newDiscordRequest(String route, String method) throws IOException {
-            Request req = new Request(!route.startsWith("http") ? "https://discord.com/api/v9" + route : route, method);
+            Request req = new Request(!route.startsWith("http") ? "https://discord.com/api/v9" + route : route, method.toUpperCase());
```
Vendicated, line 259: the request constructor already makes it lower case

scrazzz: Ok done

PR #157, opened 2021-11-23
feat(Plugin): new lifecycle methods

DiamondMiner88:
time for an overhaul

- Deprecate `load(context)` in favor of new `onLoad()`
- Deprecate `unload(context)` in favor of new `onUnload()`
- Deprecate `start(context)` in favor of new `onStart()`
- Deprecate `stop(context)` in favor of new `onStop()`
- New `onInstall()`
- New `onFirstInstall()`
- New `onBeforeUpdate(nextVersion)`
- New `onAfterUpdate(prevVersion)`
- New `onUninstall()`


Aliucord/src/main/java/com/aliucord/entities/Plugin.java
```diff
@@ -168,28 +169,104 @@ public AliucordPlugin getAnnotation() {
...
+     */
+    @SuppressWarnings("RedundantThrows")
+    public void onInstall() throws Throwable {}
+
+    /**
+     * Called when your plugin is installed for the first time
+     * Called after {@link Plugin#onInstall()}
+     */
+    @SuppressWarnings("RedundantThrows")
+    public void onFirstInstall() throws Throwable {}
```
Vendicated, line 249: What would be the use case for this?

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -239,20 +239,26 @@ public void onToggleClick(ViewHolder holder, boolean state, int position) {
...
-                PluginManager.stopPlugin(p.getName());
-                PluginManager.plugins.remove(p.getName());
-                PluginManager.logger.info(ctx, "Successfully deleted " + p.getName());
+                try {
+                    p.onUninstall();
+                } catch (Throwable err) { PluginManager.logger.error("Exception while uninstalling plugin: " + name, err); }
+
+                PluginManager.stopPlugin(name);
+                PluginManager.plugins.remove(name);
+                SettingsUtils.delete(PluginManager.getPluginPrefKey(name));
```
Vendicated, line 260: why did you add this?

Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
```diff
@@ -170,6 +174,10 @@ public static boolean update(String plugin) throws Throwable {
             );
         }
 
+        try {
+            p.onAfterUpdate(p.getManifest().version);
```
Vendicated, line 178: i think this should be called after the remount so it uses the method of the newly updated plugin instead of the old one

Vendicated:
solve conflicts please

Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
DiamondMiner88: Issue is that it might not always be triggered then (for example when the plugin is disabled when updated)

Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
```diff
@@ -174,17 +174,24 @@ public static boolean update(String plugin) throws Throwable {
             );
         }
 
-        try {
-            p.onAfterUpdate(p.getManifest().version);
-        } catch (Throwable e) { logger.error("Exception while updating plugin: " + p.getName(), e); }
+        Utils.mainThread.post(() -> {
+            if (PluginManager.isPluginEnabled(plugin))
+                PluginManager.disablePlugin(plugin);
```
Vendicated, line 179: ??

Vendicated: why do you disable the plugin ???

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
DiamondMiner88: When deleting disabled plugins and reinstalling later it's annoying to have to go enable it again

ItzOnlyAnimal: wouldn't this remove it's setting also though?

Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
DiamondMiner88: Fixed now, I think

Aliucord/src/main/java/com/aliucord/PluginManager.java
```diff
@@ -152,9 +153,15 @@ public static void togglePlugin(String name) {
...
         logger.info("Starting plugin: " + name);
-         try {
-            Objects.requireNonNull(plugins.get(name)).start(Utils.getAppContext());
-         } catch (Throwable e) { logger.error("Exception while starting plugin: " + name, e); }
+        try {
+            Plugin p = Objects.requireNonNull(plugins.get(name));
+            p.onStart();
+            if (!SettingsUtils.exists(getPluginPrefKey(name))) {
+                SettingsUtils.setBool(getPluginPrefKey(name), true);
+                p.onInstall();
```
Vendicated, line 161: only called on first install, should be called on all installs

Aliucord/src/main/java/com/aliucord/entities/Plugin.java
```diff
@@ -182,28 +184,104 @@ public AliucordPlugin getAnnotation() {
...
-    public abstract void stop(Context context) throws Throwable;
+    @Deprecated
+    @SuppressWarnings("RedundantThrows")
+    public void stop(Context context) throws Throwable {}
+
+    /**
+     * Called when your plugin is stopped
+     */
+    @SuppressWarnings("deprecation")
+    public void onStop() throws Throwable {
```
Vendicated, line 248: should be abstract to enforce implementing this method

Vendicated (changes requested):

PR #162, opened 2021-11-28
feat: improve logger & kotlin

DiamondMiner88:
- Migrate to kotlin
- Remove `[Aliucord]` in front of everything (remove tag, default module is `Aliucord` now)
- rename toast methods


Injector/src/main/java/com/aliucord/injector/Injector.java
```diff
@@ -31,7 +31,7 @@
 
 @SuppressWarnings({"ResultOfMethodCallIgnored", "JavaReflectionMemberAccess"})
 public final class Injector {
-    public static final String LOG_TAG = "Aliucord Injector";
+    public static final String LOG_TAG = "Injector";
```
Vendicated, line 34: why

PR #163, opened 2021-11-29
Fix NPE when accessing logger in PatcherExtensions

zt64:
Fixes an issue with PatcherExtensions which had an NPE when trying to access the logger field of the patcher.


Vendicated:
`name` is deprecated

Vendicated:
moving these fields inside init seems like a bad idea. Some plugins may have something like
```kt
class MyPlugin : Plugin() {
    val patches = Patches(patcher)
}
```
With your changes this would lead to NPE

zt64:
I'm unsure how else to do it without using the deprecated `name` property then

Vendicated:
you can't

zt64:
> you can't

<:guh:880053575340933151>

wingio:
<:trollely:916000880724172860>

PR #166, opened 2021-12-05
Add settings path constant

scrazzz:


Vendicated:
actually use it in other places then, otherwise this is useless

scrazzz:
ok

Juby210 (approved):

PR #168, opened 2021-12-06
Removed token login

mantikafasi:
because it was usuless


fym35:
It's not useless xd

mantikafasi:
But theres no reason to keep this as core plugin

Juby210:
There is a reason: how do you want to find the plugin if you are not logged in

mantikafasi:
:husk:

PR #169, opened 2021-12-08
Add Welcome Screen

ItzOnlyAnimal:
This welcome screen (or welcome page to be more specific) will show after the user first logs into their account

- Show how to install plugins, themes, and to check prd's pins
- Warn that there is no free nitro, and plugin downloader is built in


Aliucord/src/main/java/com/aliucord/coreplugins/welcomepage/WelcomePage.kt
```diff
@@ -0,0 +1,114 @@
...
+                typeface = ResourcesCompat.getFont(ctx, Constants.Fonts.whitney_bold)
+                setTextColor(ColorCompat.getColor(ctx, R.c.uikit_btn_bg_color_selector_red))
+                textSize = 16f
+                setOnClickListener {
+                    TransitionManager.beginDelayedTransition(fromYT)
+                    if (fromYT.visibility != View.VISIBLE) {
+                        fromYT.visibility = View.VISIBLE
+                    } else {
+                        fromYT.visibility = View.GONE
+                    }
```
zt64, line 99: These lines can be simplified to just:
```kt
fromYT.visibility = if (fromYT.visibility == View.VISIBLE) View.GONE else View.VISIBLE
```

ItzOnlyAnimal:
undrafted it

Vendicated:
did you even test this, trimIndent() doesnt work

ItzOnlyAnimal:
it does tho

Vendicated:
I think this should be part of Aliucord Manager once it ends up coming out

PR #171, opened 2021-12-10
feat(SupportWarn): Use InputDialog for confirmation 

ItzOnlyAnimal:
maybe people will actually read it now

people who make commit messages like this must be masochists


Vendicated:
what if you made custom diaolog that has quiz with multiple choice:
```
> Will I get support here?
- Yes
- No

> What will happen if i ask for support here?
- I will receive help
- I will be muted
```
and only if they answer correctly they get access

ItzOnlyAnimal:
I will do this in welcome screen /manager probably

Juby210 (approved):

ItzOnlyAnimal:
> what if you made custom diaolog that has quiz with multiple choice:
> 
> ```
> > Will I get support here?
> - Yes
> - No
> 
> > What will happen if i ask for support here?
> - I will receive help
> - I will be muted
> ```
> 
> and only if they answer correctly they get access

Since manager won't be finished before RN I think the current state of this is fine

Vendicated (approved):

PR #175, opened 2021-12-16
Simplify plugin loading process

js6pak:
Closes #174
Might be supplemented by `PluginManger` and `Plugin` kotlin rewrites later, but that will probably be breaking change

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CommandHandler.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CommandHandler.kt
@@ -28,14 +28,7 @@ import com.discord.widgets.chat.list.entries.MessageEntry
 import com.discord.widgets.chat.list.sheet.WidgetApplicationCommandBottomSheetViewModel
 
 @Suppress("UNCHECKED_CAST")
-internal class CommandHandler : Plugin() {
-  init {
-    Manifest().run {
-      name = "CommandHandler"
-      initialize(this)
-    }
-  }
-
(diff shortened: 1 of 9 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/CommandHandler.kt
```diff
@@ -28,14 +28,7 @@ import com.discord.widgets.chat.list.entries.MessageEntry
...
 @Suppress("UNCHECKED_CAST")
-internal class CommandHandler : Plugin() {
-  init {
-    Manifest().run {
-      name = "CommandHandler"
-      initialize(this)
-    }
-  }
-
+internal class CommandHandler : Plugin(Manifest().apply { name = "CommandHandler" }) {
```
Vendicated, line 31:
```suggestion
internal class CommandHandler : Plugin(Manifest("CommandHandler")) {
```

Aliucord/src/main/java/com/aliucord/entities/Plugin.java
```diff
@@ -135,18 +138,27 @@ public Manifest getManifest() {
...
-    public void initialize(Manifest manifest) {
+    public Plugin(Manifest manifest) {
         if (this.manifest != null) {
-            throw new IllegalStateException("This plugin was already initialized");
+            if (manifest != null) {
+                throw new IllegalStateException("You cannot override manifest of a plugin loaded by PluginManager");
+            }
+
+            manifest = this.manifest;
+        }
```
Vendicated, line 148: shouldn't there be an else block with `this.manifest = manifest`

PR #179, opened 2021-12-18
fix: add plugin count to /plugins

DiamondMiner88:
i forgor 💀


mantikafasi (approved):
bruh i think this can be merged without thinking a second

PR #181, opened 2021-12-19
move installer builds to releases

Juby210:
Should also fix annoying issue with having rebuild Injector to make installer detect new base update.

(workflow is not tested btw :trollface:)


Vendicated:
let's do manual releases and actually use version numbers so it's easier to tell if people's installer is outdated

Vendicated:
similar to Vanced Manager I guess

PR #182, opened 2021-12-19
feat: install plugin zips from dev channel

DiamondMiner88:
Like what themer does, but for plugins

Useful for sending people plugins to test

heres a sample: https://discord.com/channels/811255666990907402/811261478875299840/908566566310907905


ItzOnlyAnimal (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginDownloader.kt
```diff
@@ -71,6 +73,18 @@ internal class PluginDownloader : Plugin() {
                                 }
                             }
                         }
+
+                        if (msg.hasAttachments()) {
```
Vendicated, line 77: i don't like this also being in Plugin links and plugin support because we ban sending plugin zips in support because thats dangerous as some random may send malicious builds of plugins, so this should only happen in plugin dev and no other channel. You can just make this its own when case since noone is gonna send a github link in plugin dev

PR #187, opened 2021-12-27
PatcherUtils

TymanWasTaken:
This adds easy way to add buttons to commonly patched areas to just make everything more organized rather than having 98340984290 patches on one method


Aliucord/src/main/java/com/aliucord/patcher/PatcherUtils.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    // Action data
+    data class ChannelAction @JvmOverloads constructor(
+        val text: String,
+        val icon: Drawable,
+        val onClick: (view: View, model: WidgetChannelsListItemChannelActions.Model) -> Any,
+        val onLongClick: ((view: View, model: WidgetChannelsListItemChannelActions.Model) -> Boolean)? = null,
+        val condition: ((model: WidgetChannelsListItemChannelActions.Model) -> Boolean)? = null
+    )
```
Vendicated, line 135: you will document all classes and fields and methods

PR #200, opened 2022-01-20
Add page to see & delete plugins that failed to load; refactor ToolbarButtons

Vendicated:


Aliucord/src/main/java/com/aliucord/widgets/FailedPluginWidget.kt
```diff
@@ -0,0 +1,105 @@
...
+            layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+
+            setBackgroundColor(ColorCompat.getThemedColor(ctx, R.b.colorBackgroundSecondaryAlt))
+            mFilenameView = TextView(ctx, null, 0, R.i.UiKit_TextView_Bold).apply {
+                gravity = Gravity.CENTER_VERTICAL
+                layoutParams = LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.MATCH_PARENT).apply {
+                    weight = 1f
+                }
+
+                text = file.name + " (${FileUtilsKt.getSizeSubtitle(file.length())})"
```
zt64, line 64: You will use string template

ItzOnlyAnimal (commented):
plugin errors still sounds weird but i have no better name lol good

zt64 (changes requested):
Instead of hard coding IDs, maybe they could all be moved into one object that contains the IDs, would make porting versions quicker and allow plugins to reuse the ID

zt64:
:trollface:

ItzOnlyAnimal:
:trollface:

PR #245, opened 2022-05-14
Switch installer to Material 3, add dynamic color

AAGaming00:
zt will finish manager NOW!


dxcvvxd:
zt is gullible

AAGaming00:
true

AAGaming00:
lmao no way this got merhed

PR #288, opened 2022-12-12
Adding an (obnoxious) reminder that THIS is against TOS

OmegaSunkey:
just in case someone forgot they are modding apps (very rare cases like https://discord.com/channels/811255666990907402/811255667469582420/1051825614077829141)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Aliucord is a modification for the Android Discord app
 </p>
 
 ## ⚠️ Important Information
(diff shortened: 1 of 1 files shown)


ErrorOliver:
Everyone knows but discord doesn't ban them (no cases of being banned)

Only if the user is malicious, they would be banned

Meqativ:
> Everyone knows but discord doesn't ban them (no cases of being banned)
> 
> Only if the user is malicious, they would be banned

sunke didn't mention that you would get banned.. it says "Use it at your own risk"

rushiiMachine (approved):

PR #291, opened 2022-12-25
Fix badges in readme

sravan1946:
The badge currently displays the issue [badges/shields#8671](https://github.com/badges/shields/issues/8671) from the source of that badge. A breaking change requires the URL we fetch the badge from to be changed. This PR fixes the badge in accordance to that issue.


Vendicated:
thanks!

PR #347, opened 2023-05-03
Refactor project

zt64:
* Fixes module name casing
* Adds libs.versions.toml for easy versioning
* Unifies config to reduce duplicate code
* Removes deprecated manifest.patch

Ideally Aliucord core API should be in it's own module, then core plugins in another module.


rushiiMachine:
renaming the projects means the artifact names and zips are now broken afaik

zt64:
> renaming the projects means the artifact names and zips are now broken afaik

aaaaaaaaaaaaaaaaaaaaaaaaaa i will weep

Juby210:
Not needed + will break a lot of things because of renaming projects

PR #349, opened 2023-05-04
[skip ci] refactor project

zt64:


Juby210:
Keep manifest.patch, it is source of precompiled patched manifest which is **still used** by Installer

zt64:
Restored

PR #381, opened 2023-09-12
Add Pronouns core plugin

TymanWasTaken:
Displays a user's set pronouns under their username

I made `userProfile` in the `RNUserProfile` class to be nullable because it is null when a user has blocked you (thanks dolfies). That should be backwards compatible, but I'm not sure any other plugins use it in the first place

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/Pronouns.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/Pronouns.kt
@@ -14,6 +14,7 @@ import com.aliucord.patcher.after
 import com.discord.utilities.view.text.SimpleDraweeSpanTextView
 import com.discord.widgets.user.profile.UserProfileHeaderView
 import com.discord.widgets.user.profile.UserProfileHeaderViewModel
+import com.lytefast.flexinput.R
 
 val sheetProfileHeaderViewId = Utils.getResId("user_sheet_profile_header_view", "id")
 val userProfileHeaderSecondaryNameViewId = Utils.getResId("user_profile_header_secondary_name", "id")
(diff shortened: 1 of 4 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/Pronouns.kt
```diff
@@ -0,0 +1,53 @@
...
+    override fun load(context: Context) {
+        patcher.after<UserProfileHeaderView>("configureSecondaryName", UserProfileHeaderViewModel.ViewState.Loaded::class.java) {
+            val state = it.args[0] as? UserProfileHeaderViewModel.ViewState.Loaded ?: return@after
+
+            val profile = state.userProfile as? RNUserProfile ?: return@after
+            val pronouns = profile.guildMemberProfile?.pronouns?.ifEmpty { null }
+                ?: profile.userProfile?.pronouns?.ifEmpty { null }
+                ?: return@after
+
+            if (id == sheetProfileHeaderViewId) {
```
rushiiMachine, line 33: you could make this a returning guard before getting all the profile data

Aliucord/src/main/java/com/aliucord/coreplugins/Pronouns.kt
```diff
@@ -0,0 +1,53 @@
...
+                val secondaryNameView = findViewById<SimpleDraweeSpanTextView>(userProfileHeaderSecondaryNameViewId)
+                val layout = secondaryNameView.parent as LinearLayout
+
+                layout.findViewById(pronounsViewId) ?: TextView(layout.context, null, 0, com.lytefast.flexinput.R.i.UiKit_TextView_Semibold).apply {
```
rushiiMachine, line 37: import `R`?

rushiiMachine (changes requested):

rushiiMachine (approved):

PR #392, opened 2023-10-17
Add GifPreviewFix core plugin

Kyiro:


Aliucord/src/main/java/com/aliucord/coreplugins/GifPreviewFix.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.net.Uri
+import com.aliucord.entities.Plugin
+import com.aliucord.patcher.after
+import com.discord.utilities.embed.EmbedResourceUtils
+
+internal class GifPreviewFix : Plugin(Manifest("GifPreviewFix")) {
+    override fun load(context: Context) {
+        patcher.after<EmbedResourceUtils>("getPreviewUrls", String::class.java, Int::class.java, Int::class.java, Boolean::class.java) {
+            // return List<String>
+            var result = (it.result as List<String>).toMutableList()
```
rushiiMachine, line 19: is the actual type an ArrayList?
if it is then you should cast directly to avoid realloc

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/GifPreviewFix.kt
Kyiro: It's `List<String>` like the comment states.

rushiiMachine: List is an interface, not an implementation
is the type assigned an ArrayList or an immutable one?

Kyiro: Oh my bad, I'll check tomorrow/later

rushiiMachine (approved):

PR #416, opened 2024-01-20
fix: remove discrim validation for updating username

zt64:
Removed the validation for discriminator, and also hide's the discriminator input entirely if the user has already been moved to the new username system.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
@@ -23,6 +23,7 @@ import com.discord.models.deserialization.gson.InboundGatewayGsonParser
 import com.discord.models.member.GuildMember
 import com.discord.models.user.CoreUser
 import com.discord.models.user.MeUser
+import com.discord.stores.StoreStream
 import com.discord.utilities.auth.`AuthUtils$createDiscriminatorInputValidator$1`
 import com.discord.utilities.icon.IconUtils
 import com.discord.utilities.user.UserUtils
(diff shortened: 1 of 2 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
```diff
@@ -150,6 +158,27 @@ fun patchDefaultAvatars() {
     )
 }
 
+fun patchUsername() {
+    Patcher.addPatch(
+        `AuthUtils$createDiscriminatorInputValidator$1`::class.java.getDeclaredMethod("getErrorMessage", TextInputLayout::class.java),
+        InsteadHook.DO_NOTHING
```
rushiiMachine, line 164: What about those still with a legacy discrim + nitro? Or can you not even change discrim at all anymore

Kyiro: Yeah you can't update your discrim anymore, it'll only get updated when your nitro runs out.

dolfies: Not true, you can still change it with Nitro

rushiiMachine: in that case this check shouldn't be disabled if the user isn't pomelo'd yet

Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
```diff
@@ -4,9 +4,12 @@
  * Licensed under the Open Software License version 3.0
  */
 
+@file:Suppress("MISSING_DEPENDENCY_SUPERCLASS")
+
```
rushiiMachine, line 8: can u remove

Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
```diff
@@ -150,6 +159,24 @@ fun patchDefaultAvatars() {
     )
 }
 
+fun patchUsername() {
+    if (StoreStream.getUsers().me.discriminator == 0) {
```
rushiiMachine, line 163: you could use an early return here instead to reduce nesting

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
zt64: But it gets rid of the error for those funny classes
I'll remove if it's a must

Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
zt64: Oh yea I'll do that

rushiiMachine (approved):

PR #272, opened 2022-10-06
fix: user settings parser race condition

DiamondMiner88:
funny discord with the parser race condition bug

PR #438, opened 2024-03-05
fix: opcode parser bug

rushiiMachine:
ref: https://github.com/Aliucord/Aliucord/pull/272#issuecomment-1977621987

this is absolutely horrid


rushiiMachine:
nvm they fixed it like 10m ago

PR #476, opened 2024-08-26
feat(SupporterBadges): rewrite & clean up

rushiiMachine:
- Use new badges API to fetch everything at once
- Cache badges to avoid refetching
- Clean up code to make it actually readable


rushiiMachine:
@wingio review inshallah

wingio:
> @wingio review inshallah

guh its 1am

Juby210 (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/badges/Utils.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.aliucord.*
+
+private val imageCache = HashMap<String, Bitmap>()
+
+/**
+ * Set a remote image to be displayed in this view.
+ * This will be cached in memory for the lifecycle of the application.
+ * As such, this should only be used for long-living and often used images.
+ */
+internal fun ImageView.setCacheableImage(url: String) {
```
Vendicated, line 15: this seems like a bad idea, do we really need this?

you should at least add a timeout, eg cache for 10 minutes then clear

Juby210: Why? It is not like they're going to change as url includes their hash, and I don't think it is going to use a lot of memory.

rushiiMachine: This is exactly what was done before
Also in the rare event someone actually sees a badge it's likely they're going to see it again soon enough.
I just don't see a point in adding complexity for something that's never going to be an issue.

Aliucord/src/main/java/com/aliucord/coreplugins/badges/BadgesAPI.kt
```diff
@@ -0,0 +1,68 @@
...
+     * Get cached badge data or re-fetch from the API if expired.
+     */
+    @OptIn(ExperimentalTime::class)
+    fun getBadges(): BadgesInfo {
+        if (!isCacheExpired()) return settings.cachedBadges
+
+        val data = fetchBadges()
+
+        return if (data != null) {
+            settings.cacheExpiration = System.currentTimeMillis() + Duration.days(1).inWholeMilliseconds
```
wingio, line 23: do we not have the extension vals for duration here? (eg 1.days)

rushiiMachine: I have no idea why, but kotlin deprecated those
it's dumb

wingio: how recently? because I've used them not that long ago and they weren't deprecated then. But also i don't think it matters for us since we can't really use the newer versions of kotlin anyways

Aliucord/src/main/java/com/aliucord/coreplugins/badges/BadgesAPI.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        val data = fetchBadges()
+
+        return if (data != null) {
+            settings.cacheExpiration = System.currentTimeMillis() + 1.days.inWholeMilliseconds
+            settings.cachedBadges = data
+            data
+        } else {
+            // Failed to fetch; keep cache and try later
+            settings.cacheExpiration = System.currentTimeMillis() + 1.days.inWholeMilliseconds
```
wingio, line 29: isn't this supposed to be 6 hours

rushiiMachine: oops

wingio (approved):

PR #505, opened 2025-01-02
ForwardedMessages

mantikafasi:
currently only supports viewing, planning to add option to jump source in widgetchatlistactions

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/ForwardedMessages.java
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/ForwardedMessages.java
@@ -1,0 +1,61 @@
+package com.aliucord.coreplugins;
+
+import android.annotation.SuppressLint;
+import android.content.Context;
+import android.view.View;
+import android.widget.TextView;
+
+import com.aliucord.entities.CorePlugin;
+import com.aliucord.patcher.Hook;
+import com.aliucord.patcher.PreHook;
+import com.aliucord.utils.ReflectUtils;
+import com.discord.api.message.Message;
+import com.discord.widgets.chat.list.adapter.WidgetChatListAdapterItemMessage;
+
+import java.util.ArrayList;
+
+public class ForwardedMessages extends CorePlugin {
+    public ForwardedMessages() {
+        super(new Manifest("ForwardedMessages"));
+    }
+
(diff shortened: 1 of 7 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/ForwardedMessages.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.ArrayList;
+
+public class ForwardedMessages extends CorePlugin {
+    public ForwardedMessages() {
+        super(new Manifest("ForwardedMessages"));
+    }
+
+    @SuppressLint("SetTextI18n")
+    @Override
+    public void start(Context context) throws Throwable {
```
rushiiMachine, line 24: You should:
1. Update the Injector & patches project version in their respective `build.gradle.kts`
2. Here, check if that version is at least that new version using the [`ManagerBuild`](https://github.com/Aliucord/Aliucord/blob/c91a6956e7b0b1808039e3bac3281e28a6c73e0f/Aliucord/src/main/java/com/aliucord/updater/ManagerBuild.kt) class, otherwise return early here.

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/ForwardedMessages.java
mantikafasi: lets just dont let old clients update to this build and force them to update using manager, adding check to every core plugin would be annoying

rushiiMachine: That moves the complexity elsewhere and prevents unrelated updates from being deployed

KraXen72:
hype 🔥🔥🔥

Aliucord/src/main/java/com/aliucord/coreplugins/ForwardedMessages.java
mantikafasi: I updated version

KraXen72:
@mantikafasi what happened :sob:

KraXen72:
ok found the new pr sorry

PR #534, opened 2025-05-25
fix(UploadSize): lower non-nitro upload limit

Canny1913:
Adjusts upload limit for non-nitro users since Discord lowered it to 10 MB.


Canny1913:
> This should be replaced with newer per-attachment size limit instead. I don't really think lowering it now changes anything as it still counts per-message.

Done. (unless i made a dumb mistake)

Juby210 (approved):

PR #541, opened 2025-06-08
Update build.gradle.kts

pikaczu787:


pikaczu787:
@Vendicated 
add back issues in vencord

PR #560, opened 2025-07-10
feat: improve bug issue template

zt64:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,59 +1,86 @@
...
+description: Report a bug or crash in Aliucord. For plugin issues, please use the plugin's repo or our support server.
 labels: [bug]
 body:
   - type: input
     id: discord
     attributes:
-      label: Discord Account
-      description: Who on Discord is making this request? Not required.
-      placeholder: username#0000
+      label: Discord username (Optional)
```
rushiiMachine, line 8: I think we should remove this altogether, it's never been useful

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,59 +1,86 @@
...
+        3. App crashes immediately
+    validations:
+      required: true
+
+  - type: input
+    id: android-version
+    attributes:
+      label: Android Version
+      description: What Android version is this happening on?
+      placeholder: Android 13 (One UI 5.1), or SDK 33
```
rushiiMachine:
```suggestion
      placeholder: Android 13, or SDK 33
```
ROM is already specified in the field below

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,59 +1,86 @@
...
+      label: Crash Log (if applicable)
+      description: Paste the crash log (between triple backticks ```).
+      placeholder: |
         ```
-        Replace this text with your crash-log.
+        Paste your crash log here
         ```
     validations:
       required: false
+
```
rushiiMachine, line 78: There should also be an optional field to provide logcats if there's no crashlog, and link this as a tutorial: https://pastebin.com/pNhXwhrd

I dunno if gh supports attachments as an entry type but if it does do that

rushiiMachine (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
Juby210: Nop, keep it, having context is better. I sometimes check if they sent something in support before.

.github/ISSUE_TEMPLATE/bug_report.yml
rushiiMachine: will you fix

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,59 +1,90 @@
...
       options:
-        - label: I did indeed check to make sure the bug or crash report is applicable!
-          required: true
\ No newline at end of file
+        - label: I’ve checked that this issue hasn’t already been reported.
+          required: true
+
+  - type: markdown
+    attributes:
+      value: Upload a file containing logcat output, see [here](https://pastebin.com/pNhXwhrd) for how
```
rushiiMachine:
```suggestion
      value: After creating this issue, please upload a file containing a logcat in a separate comment! See [here](https://pastebin.com/pNhXwhrd) on how to accomplish this.
```

rushiiMachine (changes requested):

PR #562, opened 2025-07-13
Allow zip plugin installs in #support too

Archimedes9500:
:husk:


system32149:
the greatest pull request thats ever lived

rushiiMachine:
:x: unless this is limited to those who have either the mod or plugin dev role, then no.

PR #570, opened 2025-07-24
refactor: convert java coreplugins to kotlin

zt64:
I tested and no errors occurred

This is only the beginning.


rushiiMachine:
oh the horrors

rushiiMachine:
I'll be honest this looks messier and **a lot** less readable than the old source code.
I don't think it's worth rewriting these, they're fine as is.
No point in rewriting a coreplugin no one will ever touch again

rushiiMachine:
There's places where the Java src is incredibly messy but these 3 coreplugins are not one of them
I understand if rewriting was to make the code cleaner (might as well do it in Kotlin), but rewriting to Kotlin just for the sake of it is 👎

rushiiMachine:
TLDR this does not improve the code of those coreplugins, so I'm closing this
ref: the discussion we had in discord a few weeks ago

PR #571, opened 2025-07-26
feat: add MessageWidthFix core plugin

zt64:
Adds plugin to fix message width so that the on click and on long click listeners are able to be used. This only applies to embeds and stickers afaik. Unfortunately the ripple effect isn't present, there seems to be a selectable background element that the embeds and sticker items lack.


rushiiMachine:
any update on this

PR #575, opened 2025-08-02
feat: stop crashlytics folder from being made on startup

omardotdev:
i made smali patches which stops  `.com.google.firebase.crashlytics` and `.com.google.firebase.crashlytics-ndk` from being made so they don't add an additional 16kB on startup which overtime can increase the user data size by alot

### DIFF
--- a/patches/src/b/i/c/m/e/b.patch
+++ b/patches/src/b/i/c/m/e/b.patch
@@ -1,0 +1,25 @@
+--- smali_original/b/i/c/m/e/b.smali
++++ smali/b/i/c/m/e/b.smali
+@@ -41,16 +41,16 @@ .method public a(Lb/i/c/l/e;)Ljava/lang/
+     check-cast p1, Landroid/content/Context;
+ 
+     .line 3
+-    new-instance v0, Ljava/io/File;
+-
+-    invoke-virtual {p1}, Landroid/content/Context;->getFilesDir()Ljava/io/File;
++    :cond_1a
++    sget-object v0, Lb/i/c/m/d/b;->a:Lb/i/c/m/d/b;
+ 
+-    move-result-object v1
++    const-string v1, "Couldn\'t create file"
+ 
+-    const-string v2, ".com.google.firebase.crashlytics-ndk"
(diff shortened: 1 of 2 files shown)


patches/src/b/i/c/m/e/b.patch
rushiiMachine, line 1: I'm trying to understand what this is supposed to do. Could you explain a bit?

omaawr: both patches make the file so instead of making the crashlytics folder, it prints "Couldn't read file" and returns null which avoids the folder from being made

rushiiMachine: Prints where though? Whats the purpose of doing it at all?

omaawr: logs maybe? i think i might remove that since well, its completely useless but i dont have much time currently..

rushiiMachine: yeah please remove it

omaawr: okay i've removed it, sorry for the delay

patches/src/b/i/c/m/e/b.patch
```diff
@@ -0,0 +1,23 @@
...
+     .line 3
+-    new-instance v0, Ljava/io/File;
+-
+-    invoke-virtual {p1}, Landroid/content/Context;->getFilesDir()Ljava/io/File;
+-
+-    move-result-object v1
+-
+-    const-string v2, ".com.google.firebase.crashlytics-ndk"
++    :cond_1a
++    sget-object v0, Lb/i/c/m/d/b;->a:Lb/i/c/m/d/b;
```
rushiiMachine: I'm fairly sure you can just remove this `sget-object` and retain the label, as you immediately replace the value set to `v0` with null on the next instruction

rushiiMachine (changes requested):

patches/src/b/i/c/m/e/b.patch
omaawr: i think i've fixed it probably? not entirely sure..

PR #581, opened 2025-08-16
fix(Polls): fix various bugs

cillynder:
- Fixes a crash in details screen, which can happens when an uncached user's id is added from the poll add gateway event.
- Fixes a loop where the user will get stuck in the poll creation screen if they swiped to it.
- Adds a character limit to question and answer texts when creating polls.

### DIFF
--- a/Aliucord/build.gradle.kts
+++ b/Aliucord/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = "com.aliucord"
-version = "2.3.0"
+version = "2.3.1"
 
 aliucord {
     projectType.set(com.aliucord.gradle.ProjectType.CORE)


Juby210 (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/polls/details/PollDetailsScreen.kt
```diff
@@ -108,20 +117,21 @@ internal class PollDetailsScreen : AppFragment(Utils.getResId("widget_manage_rea
...
+                val userItems = snapshot.voters.mapNotNull { id ->
+                    usersMap[id]?.let { user ->
+                        ManageReactionsResultsAdapter.ReactionUserItem(
+                            user,
+                            0,
+                            0,
+                            MessageReactionEmoji("", "", false),
+                            false,
+                            membersMap?.get(id)
+                        )
```
rushiiMachine: While you're at it, could you add parameter names as comments here?

Aliucord/src/main/java/com/aliucord/coreplugins/polls/details/PollDetailsScreen.kt
```diff
@@ -75,18 +77,25 @@ internal class PollDetailsScreen : AppFragment(Utils.getResId("widget_manage_rea
         setActionBarTitle("Poll Votes")
         setActionBarSubtitle(poll.question.text)
 
-        subscription = PollsStore.subscribeOnMain(channelId, messageId) {
-            data = it
-            updateData()
+        subscription = PollsStore.subscribeOnMain(channelId, messageId) { newData ->
+            if (newData == null) { // Message is deleted
+                this.appActivity.finish()
+                return@subscribeOnMain
```
rushiiMachine, line 83: Shouldn't you unsubscribe here?

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/polls/details/PollDetailsScreen.kt
cillynder: The system will call `onDestroy()` for us, where it will get unsubscribed.

rushiiMachine (approved):

rushiiMachine:
oh btw you wanna update the core version as well

PR #582, opened 2025-08-21
feat(Patches): add `.gif` sticker format

Archimedes9500:
Smali patching in GIF formatType breaks things that work fine with current `Patches.kt`, and the only feature that's actually missing is the ability to send links to gif stickers, so I patched the fallback `""` format extension to be `".gif"` instead since there is no more sticker types anyway.


rushiiMachine:
@Juby210 what do you think

Juby210:
merge

PR #586, opened 2025-08-24
feat: support new pins features

LavaDesu:
Adds support for the new pins permission, which will soon be the only permission to determine whether or not a user can pin messages (manage messages will not allow users to pin messages anymore).

Also adds support for reading more than 50 pins, as the limit was increased to 250.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
@@ -2,20 +2,30 @@ package com.aliucord.coreplugins
 
 import android.content.Context
 import android.view.View
-import android.view.ViewGroup.LayoutParams.MATCH_PARENT
-import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.LinearLayout
+import android.widget.LinearLayout.LayoutParams
+import com.aliucord.Http
 import com.aliucord.Utils
 import com.aliucord.entities.CorePlugin
 import com.aliucord.patcher.*
+import com.aliucord.utils.GsonUtils
+import com.aliucord.utils.SerializedName
 import com.aliucord.utils.ViewUtils.addTo
 import com.discord.models.message.Message
 import com.discord.models.user.MeUser
+import com.discord.stores.*
 import com.discord.utilities.permissions.ManageMessageContext
 import com.discord.utilities.permissions.PermissionUtils
+import com.discord.utilities.rest.RestAPI
 import com.discord.views.TernaryCheckBox
 import com.discord.widgets.channels.permissions.WidgetChannelSettingsEditPermissions
 import com.discord.widgets.channels.permissions.`WidgetChannelSettingsEditPermissions$permissionCheckboxes$2`
+import com.discord.widgets.chat.list.adapter.WidgetChatListAdapter
+import com.discord.widgets.chat.list.entries.LoadingEntry
+import com.discord.widgets.chat.pins.WidgetChannelPinnedMessages
+import rx.Observable
+import java.net.URLEncoder
+import java.util.WeakHashMap
 
 internal class NewPins : CorePlugin(Manifest("NewPins")) {
     override val isHidden: Boolean = true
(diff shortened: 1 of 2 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+            param.setResult(
+                ManageMessageContext(
+                    result.canManageMessages,
+                    result.canEdit,
+                    result.canDelete,
+                    result.canAddReactions,
+                    canPin,
+                    result.canMarkUnread
+                )
+            )
```
zt64:
```kt
            param.result = ManageMessageContext(
                result.canManageMessages,
                result.canEdit,
                result.canDelete,
                result.canAddReactions,
                canPin,
                result.canMarkUnread
            )
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+import java.net.URLEncoder
+import java.util.WeakHashMap
+
+internal class NewPins : CorePlugin(Manifest("NewPins")) {
+    override val isHidden: Boolean = true
+    override val isRequired: Boolean = true
+
+    companion object {
+        const val PIN_MESSAGES_PERMISSION = 1L shl 51
+    }
```
zt64: This could be private I think
```kt
    private companion object {
        const val PIN_MESSAGES_PERMISSION = 1L shl 51
    }
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+                @Suppress("UNCHECKED_CAST")
+                val lastPins = StorePinnedMessages.`access$getPinnedMessages$p`(pinStore)[adapter.data.channelId] as List<Message>
+                StorePinnedMessages.`access$handlePinnedMessagesLoaded`(
+                    StoreStream.getPinnedMessages(),
+                    adapter.data.channelId,
+                    lastPins + olderPins,
+                )
+                adapter.extension.isFetching = false
+            }
+        }
```
zt64: It might be better for readability to use a try..catch..finally statement here

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+
+        // Adds a cute loading icon when there's more to be loaded
+        patcher.before<WidgetChannelPinnedMessages>(
+            "configureUI",
+            WidgetChannelPinnedMessages.Model::class.java
+        ) { (param, model: WidgetChannelPinnedMessages.Model) ->
+            val adapter = widgetAdapterMap[this]
+                ?: return@before
```
zt64:
```kt
            val adapter = widgetAdapterMap[this] ?: return@before
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    private data class GetChannelPinsResponse(
+        val items: List<MessagePin>,
+        @SerializedName("has_more") val hasMore: Boolean,
+    ) {
+        data class MessagePin(
+            @SerializedName("pinned_at") val pinnedAt: String,
+            val message: com.discord.api.message.Message,
+        )
+    }
```
zt64:
```kt
    private data class GetChannelPinsResponse(
        val items: List<MessagePin>,
        @SerializedName("has_more") 
        val hasMore: Boolean,
    ) {
        data class MessagePin(
            @SerializedName("pinned_at") 
            val pinnedAt: String,
            val message: ApiMessage,
        )
    }
```

And add this import:
```kt
import com.discord.api.message.Message as ApiMessage
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+            val manageMessagesCheckboxId = Utils.getResId("channel_permission_text_manage_messages", "id")
+            val manageMessagesCheckbox: TernaryCheckBox = view.findViewById(manageMessagesCheckboxId)
+            val index = layout.indexOfChild(manageMessagesCheckbox) + 1
+
+            // TODO: Should be changed after 2026/01/12, or when official client does
+            manageMessagesCheckbox.setLabel("Manage Messages ⚠")
+            manageMessagesCheckbox.setSubtext(
+                "Members with this permission can delete messages by other members or pin any message.*\n\n" +
+                "* Pinning messages now has a separate permission. This setting's behaviour will change soon."
+            )
```
zt64: Could you use the setter instead of the method, or is it not generated for these fields? I'm not sure if it's generated or not

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        // Patches logic that handles configuring initial state of checkboxes
+        patcher.after<`WidgetChannelSettingsEditPermissions$permissionCheckboxes$2`>("invoke") { param ->
+            val widget = this.`this$0`
+            val view = widget.view
+            if (view == null) {
+                logger.error(IllegalStateException("View was not initialised"))
+                return@after
+            }
```
zt64, line 398:
```kt
            val view = this.`this$0`.view
            if (view == null) {
                logger.error(IllegalStateException("View was not initialized"))
                return@after
            }
```

zt64 (changes requested):
Looks mostly alright. Just some improvements in formatting can be made to improve readability

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
Vendicated: why's this even a companion can't this just be a top level const

rushiiMachine: there's no difference really
top level consts are singleton classes anyways
and the const is inlined

wingio: i think being a part of a companion better communicates ownership

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
rushiiMachine: Since `this$0` is unclear then it's better to leave this as-is

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
cillynder: There's no getters for them so Kotlin won't do its magic here

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
cillynder: Yeah I put it in companion object so accessing it would require going through NewPins. I'll just make it private too.

cillynder: Actually since it's private anyway might as well move it out

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,289 @@
...
+                isPrivateChannel: Boolean,
+                isSystemDM: Boolean,
+                isArchivedThread: Boolean,
+            ) ->
+            val result = param.result as ManageMessageContext
+
+            // TODO: After 2026/01/12, users will no longer be able to pin messages with the
+            // Manage Messages permission, and this line should be removed around the same time.
+            // https://discord.com/developers/docs/change-log#pin-permission-split
+            if (result.canTogglePinned) return@after
```
rushiiMachine: I think you should check the time here and do it, we're going to likely forget about this (and the users might not update). Keep the TODO though

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,289 @@
...
+            "onViewBound",
+            View::class.java,
+        ) { (_, view: View) ->
+            val layoutId = Utils.getResId("channel_permissions_text_container", "id")
+            val layout: LinearLayout = view.findViewById(layoutId)
+            val manageMessagesCheckboxId = Utils.getResId("channel_permission_text_manage_messages", "id")
+            val manageMessagesCheckbox: TernaryCheckBox = view.findViewById(manageMessagesCheckboxId)
+            val index = layout.indexOfChild(manageMessagesCheckbox) + 1
+
+            // TODO: Should be changed after 2026/01/12, or when official client does
```
rushiiMachine: Same thing here

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,289 @@
...
+        val pinMessagesCheckboxViewId = View.generateViewId()
+
+        // Patches the editor to add our new pin messages checkbox and change the information
+        // regarding the manage messages permission
+        patcher.before<WidgetChannelSettingsEditPermissions>(
+            "onViewBound",
+            View::class.java,
+        ) { (_, view: View) ->
+            val layoutId = Utils.getResId("channel_permissions_text_container", "id")
+            val layout: LinearLayout = view.findViewById(layoutId)
```
rushiiMachine:
```suggestion
            val layout = view.findViewById<LinearLayout>(layoutId)
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,289 @@
...
+            val manageMessagesCheckboxId = Utils.getResId("channel_permission_text_manage_messages", "id")
+            val manageMessagesCheckbox: TernaryCheckBox = view.findViewById(manageMessagesCheckboxId)
```
rushiiMachine:
```suggestion
            val manageMessagesCheckbox = view.findViewById<TernaryCheckBox>(manageMessagesCheckboxId)
```

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,289 @@
...
+        // Patches logic that handles configuring initial state of checkboxes
+        patcher.after<`WidgetChannelSettingsEditPermissions$permissionCheckboxes$2`>("invoke") { param ->
+            val widget = this.`this$0`
+            val view = widget.view
+            if (view == null) {
+                logger.error(IllegalStateException("View was not initialised"))
+                return@after
+            }
+
+            val pinMessagesCheckbox: TernaryCheckBox? = view.findViewById(pinMessagesCheckboxViewId)
```
rushiiMachine:
```suggestion
            val pinMessagesCheckbox = view.findViewById<TernaryCheckBox?>(pinMessagesCheckboxViewId)
```
not sure about the null here but try anyways

cillynder:
I somehow completely forgot about role permissions so this PR isn't completed yet TwT (only the channel permissions are implemented).

Aliucord/src/main/java/com/aliucord/coreplugins/NewPins.kt
```diff
@@ -0,0 +1,420 @@
...
+
+        // Adds a cute loading icon when there's more to be loaded
+        patcher.before<WidgetChannelPinnedMessages>(
+            "configureUI",
+            WidgetChannelPinnedMessages.Model::class.java
+        ) { (param, model: WidgetChannelPinnedMessages.Model) ->
+            val adapter = widgetAdapterMap[this] ?: return@before
+
+            if (adapter.extension.hasMore) {
+                param.args[0] = model.copy(
```
rushiiMachine: Is this really the only way? I'm pretty hesitant about this much unnecessary copying.
Would it be possible to just set the list of items on the existing model via reflection (and if it's an ArrayList internally you could just append)?

cillynder: Sure reflection is possible! I just wanted to avoid using it since I thought it was just a shallow data class copy.

rushiiMachine (approved):

PR #593, opened 2025-08-29
feat(Decorations): init

LavaDesu:
Lays the groundwork for supporting user decorations. The core plugin will be disabled until it's ready (currently no user-facing features yet).

- Also adds some more view utils.
- Also adds some really useful reflection delegates.
- Bumps Injector to 2.3.0, patches to 1.3.0.

Some signatures of `SettingsDelegates.kt` had to be changed but I tried to make sure they are backwards-compatible; however, `SettingsDelegate` is not. But I assume there aren't any plugins using that instead of `SettingsAPI.delegate`.


rushiiMachine:
whats fluff stand for loll

cillynder:
> whats fluff stand for loll

It's supposed to mean unnecessary features x,) but the name can be changed. My idea was for this one coreplugin to handle all the different kinds of decorations that not every user wants, and by putting them in one place users can more easily choose to hide what they want.

Aliucord/src/main/java/com/aliucord/utils/ReflectDelegates.kt
```diff
@@ -18,14 +20,104 @@ import kotlin.reflect.KProperty
...
+ *
+ * @param clazz     The class that the method belongs to.
+ * @param methodName The name of the method.
+ */
+class LazyMethod<T>(private val clazz: Class<*>, private val methodName: String?) : ReadOnlyProperty<T, Method> {
+    private var v = null as Method?
+    override fun getValue(thisRef: T, property: KProperty<*>) = v ?: clazz.getDeclaredMethod(methodName ?: property.name).apply {
+        isAccessible = true
+        v = this
+    }
```
rushiiMachine: long expression function bodies like this are super hard to read
you should make this a regular multiline function

Aliucord/src/main/java/com/aliucord/utils/ReflectDelegates.kt
```diff
@@ -18,14 +20,104 @@ import kotlin.reflect.KProperty
...
+ * @param clazz The class holding the field.
+ * @param fieldName The name of the field. If null, will use the property name.
+ */
+class FieldAccessor<T, U>(private val clazz: Class<T>, private val fieldName: String?): ReadWriteProperty<T, U> {
+    private var field: Field? = null
+    private fun field(property: KProperty<*>): Field {
+        field?.let { return it }
+
+        return clazz.getDeclaredField(
+            fieldName ?: property.name.replace("Field", "")
```
rushiiMachine: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/remove-prefix.html

Aliucord/src/main/java/com/aliucord/utils/DimenUtils.kt
rushiiMachine, line 1: While you're at it, could you go and make all of these inline? The `density` field will have to become internal and be marked with `@PublishedApi` I think

Aliucord/src/main/java/com/aliucord/coreplugins/decorations/DecorationsSettings.kt
```diff
@@ -0,0 +1,38 @@
...
+                CheckedSetting.ViewType.SWITCH,
+                description,
+                null
+            ).apply {
+                var setting by delegate
+                isChecked = setting
+                setOnCheckedListener { setting = !setting }
+            }
+        }
+    }
```
rushiiMachine, line 37: Is this groundwork for toggling individual profile decor things?

rushiiMachine:
> > whats fluff stand for loll
> 
> It's supposed to mean unnecessary features x,) but the name can be changed. My idea was for this one coreplugin to handle all the different kinds of decorations that not every user wants, and by putting them in one place users can more easily choose to hide what they want.

ah i see
I do think the name should be changed to something that actually describes it though. ie, `ProfileDecor` 
Don't forget the package naming

Aliucord/src/main/java/com/aliucord/coreplugins/decorations/DecorationsSettings.kt
cillynder: yep! it'll look something like [this](https://github.com/Aliucord/Aliucord/pull/594/commits/c845e8ce70dea46e0520f3806b969b0bf7e35b8e#diff-2489826cb7933117bdbef6695caecc7b3aee58e666af450549ab50a85bb9d4f9)

Aliucord/src/main/java/com/aliucord/utils/DimenUtils.kt
cillynder: I could inline the extension properties, but the static functions it would complain about it not being necessary

Also, I think it would break compatibility for plugins that use it(?) since the functions would disappear

rushiiMachine: > the static functions it would complain about it not being necessary

You can just suppress.

> it would break compatibility for plugins

I believe inlined functions do not get omitted after inlining, but you could decompile and double check

Aliucord/src/main/java/com/aliucord/coreplugins/Decorations.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.discord.api.user.User
+import com.discord.models.member.GuildMember
+import com.discord.models.user.CoreUser
+import com.discord.models.user.MeUser
+import com.discord.stores.StoreGuilds
+import com.discord.api.guildmember.GuildMember as ApiGuildMember
+
+internal class Decorations : CorePlugin(Manifest().apply {
+    name = "Decorations"
+    description = "Adds support for various user decorations"
```
rushiiMachine:
```suggestion
    description = "Adds support for various user profile decorations"
```

Aliucord/src/main/java/com/aliucord/coreplugins/Decorations.kt
```diff
@@ -0,0 +1,94 @@
...
+            displayNameStyles = api.displayNameStyles
+            primaryGuild = api.primaryGuild
+        }
+        patcher.after<CoreUser.Companion>("merge", CoreUser::class.java, User::class.java) { (param, old: CoreUser, api: User) ->
+            val res = param.result as CoreUser
+
+            (api.avatarDecorationData ?: old.avatarDecorationData)?.let { res.avatarDecorationData = it }
+            (api.collectibles ?: old.collectibles)?.let { res.collectibles = it }
+            (api.displayNameStyles ?: old.displayNameStyles)?.let { res.displayNameStyles = it }
+            (api.primaryGuild ?: old.primaryGuild)?.let { res.primaryGuild = it }
```
rushiiMachine, line 54: What was the point of making each assignment conditional? Is this how it does the same for other values internally?

Aliucord/src/main/java/com/aliucord/coreplugins/Decorations.kt
```diff
@@ -0,0 +1,94 @@
...
+            displayNameStyles = api.displayNameStyles
+            primaryGuild = api.primaryGuild
+        }
+        patcher.after<MeUser.Companion>("merge", MeUser::class.java, User::class.java) { (param, old: MeUser, api: User) ->
+            val res = param.result as MeUser
+
+            (api.avatarDecorationData ?: old.avatarDecorationData)?.let { res.avatarDecorationData = it }
+            (api.collectibles ?: old.collectibles)?.let { res.collectibles = it }
+            (api.displayNameStyles ?: old.displayNameStyles)?.let { res.displayNameStyles = it }
+            (api.primaryGuild ?: old.primaryGuild)?.let { res.primaryGuild = it }
```
rushiiMachine, line 69: Same here

rushiiMachine:
Oh also could you bump the core version in this PR as well

Aliucord/src/main/java/com/aliucord/coreplugins/Decorations.kt
cillynder: The idea of the merge method is that a new, possibly partial user api object (second argument) can get merged with an existing cached one (first argument). So we use the new one if it exists, otherwise fall back on the cached one. Internally, they uses their own NullSerializable object which provides a way to distinguish if a field is null or missing, which is technically more correct.

I haven't done enough testing to check in what places this actually gets called, but the possible worst case scenario rn is that someone removing a decoration won't be immediately reflected.

I'll do some testing and report back.

rushiiMachine: What I'm trying to say is that if the result value of that method is a copy of `old` with updated fields, what's the difference between
```kt
            (api.avatarDecorationData ?: old.avatarDecorationData)?.let { res.avatarDecorationData = it }
```
and
```kt
            res.avatarDecorationData = api.avatarDecorationData ?: old.avatarDecorationData
```
Although now that I think about it, if it were null then skipping assignment would actually save some time.

cillynder: You're right; there isn't much difference between the two to be honest, since res is a new object and will always have those fields as null in the first place. I think in some other places this isn't the case (res is a copy rather than a new instance), which is why I did it like this to avoid overwriting it with null.

PR #595, opened 2025-08-29
feat: remove billing

omardotdev:
i made it so billing and links to it is removed including settings, the boosting page and also the gift button

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/PluginManager.kt
+++ b/Aliucord/src/main/java/com/aliucord/PluginManager.kt
@@ -270,6 +270,7 @@ object PluginManager {
     @JvmStatic
     fun loadCorePlugins(context: Context) {
         val corePlugins = arrayOf(
+            AlignThreads(),
             ButtonsAPI(),
             CommandHandler(),
             CoreCommands(),
@@ -283,17 +284,16 @@ object PluginManager {
             Polls(),
             PrivateChannelsListScroll(),
             PrivateThreads(),
-            RestartButton(),
-            RemoveBilling(),
-            RNAPI(),
             Pronouns(),
+            RNAPI(),
+            RemoveBilling(),
+            RestartButton(),
             ShowReplyMention(),
             StickerCrashFix(),
             SupportWarn(),
             SupporterBadges(),
             TokenLogin(),
             UploadSize(),
-            AlignThreads(),
         )
 
         corePlugins.forEach { p ->
--- a/Aliucord/src/main/java/com/aliucord/PluginManager.kt
+++ b/Aliucord/src/main/java/com/aliucord/PluginManager.kt
@@ -284,6 +284,7 @@ object PluginManager {
             PrivateChannelsListScroll(),
             PrivateThreads(),
             RestartButton(),
+            RemoveBilling(),
             RNAPI(),
             Pronouns(),
             ShowReplyMention(),
@@ -293,7 +294,6 @@ object PluginManager {
             TokenLogin(),
             UploadSize(),
             AlignThreads(),
-            RemoveBilling()
         )
 
         corePlugins.forEach { p ->
--- a/Aliucord/src/main/java/com/aliucord/PluginManager.kt
+++ b/Aliucord/src/main/java/com/aliucord/PluginManager.kt
@@ -292,7 +292,8 @@ object PluginManager {
             SupporterBadges(),
             TokenLogin(),
             UploadSize(),
-            AlignThreads()
+            AlignThreads(),
+            RemoveBilling()
         )
 
         corePlugins.forEach { p ->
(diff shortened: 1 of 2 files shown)


system32149:
so youre telling me that to boost a server using the 2 free boosts from nitro, i have to use RN??? :husk:

omaawr:
> so youre telling me that to boost a server using the 2 free boosts from nitro, i have to use RN??? :husk:

guhh you can just disable the plugin or use a computer

Aliucord/src/main/java/com/aliucord/PluginManager.kt
```diff
@@ -292,7 +292,8 @@ object PluginManager {
             SupporterBadges(),
             TokenLogin(),
             UploadSize(),
-            AlignThreads()
+            AlignThreads(),
+            RemoveBilling()
```
rushiiMachine: Could you re-alphabetize this coreplugins list? Also add a trailing comma

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -0,0 +1,88 @@
...
+import b.a.a.b.d
+import b.a.a.b.c
+import com.discord.widgets.stickers.UnsendableStickerPremiumUpsellDialog
+import com.discord.widgets.settings.WidgetSettingsMedia
+
+internal class RemoveBilling : CorePlugin(Manifest("RemoveBilling")) {
+    override val isHidden = true
+    
+    init {
+        manifest.description = "Removes all references to billing"
```
```suggestion
        manifest.description = "Removes all references to billing and nitro"
```

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    override fun start(context: Context) {
+        // Remove Billing settings
+        patcher.after<WidgetSettings>("onViewBound", View::class.java) { (_, view: View) ->
+            var container: LinearLayout = view.findViewById(Utils.getResId("nitro_settings_container", "id"))
```
```suggestion
            var container = view.findViewById<LinearLayout>(Utils.getResId("nitro_settings_container", "id"))
```
I avoid defining the variable type rather than passing it directly to the reified function, it's unclear.
Also do this for any other instances of this happening in your changes.

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -0,0 +1,88 @@
...
+            var ids = arrayOf("settings_nitro", "nitro_boosting", "nitro_header")
+            for (i in ids) {
+                container.removeView(view.findViewById(Utils.getResId(i, "id")))
```
```suggestion
            for (id in arrayOf("settings_nitro", "nitro_boosting", "nitro_header")) {
                container.removeView(view.findViewById(Utils.getResId(id, "id")))
```

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -0,0 +1,88 @@
...
+            removeTextView(view, Utils.getResId("compression_toggle_subtext", "id"))
+        }
+    }
+
+    override fun stop(context: Context) = patcher.unpatchAll()
+
+    private fun removeTextView(rootView: View, textViewId: Int) {
+        val textView = rootView.findViewById<TextView>(textViewId)
+        textView.setVisibility(View.GONE)
+    }
```
rushiiMachine: There's not really a need to do something like this. The additional cognitive load tiny utility functions like these add is not worth the few lines of code it will save. Just inline each usage

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
rushiiMachine, line 1: Also please fix the new kotlinc warnings about name shadowing

rushiiMachine (changes requested):

rushiiMachine:
Also is there any other nitro upsells still present after this?

omaawr:
> Also is there any other nitro upsells still present after this?

there was one that i forgot to do which was the learn more button whenever you tried to upload an animated avatar but i'll quickly patch it right now

Aliucord/src/main/java/com/aliucord/PluginManager.kt
rushiiMachine: They're still not alphabetized! AlignThreads is at the bottom and RemoveBilling is below RestartButton

Just get a IDE extension to sort lines and use that

omaawr: but it already is?

rushiiMachine: It is not
https://github.com/Aliucord/Aliucord/pull/595/files

omaawr: ohh wait im stupid

rushiiMachine:
we should still remove the gift button from chat bar
it's ui clutter that shouldn't be there

omaawr:
> we should still remove the gift button from chat bar it's ui clutter that shouldn't be there

gifting still actually works though.. are you sure about this?

rushiiMachine:
yes, it's junk

PR #596, opened 2025-08-30
fix(PluginDownloader): duplicate install buttons in plugin development channel

omardotdev:
for some reason the code checks twice if a message has a plugin zip file so as a result, there would be 2 buttons which are "Install {plugin}" and "Install {plugin}.zip"..


rushiiMachine:
this might've been test code I forgot to remove lmao

rushiiMachine (approved):

omaawr:
> this might've been test code I forgot to remove lmao

oh lol

PR #597, opened 2025-09-01
fix(PluginDownloader): add check to prevent from installing a plugin that has the same name as a coreplugin

omardotdev:


Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
```diff
@@ -21,7 +22,11 @@ internal class PluginFile(val plugin: String) : File("${Constants.PLUGINS_PATH}/
     fun install(url: String, callback: Runnable? = null) {
         Utils.threadPool.execute {
             val isReinstall = isInstalled
-
+            if (PluginManager.plugins[plugin] is CorePlugin) {
+                Utils.showToast("You cannot install this plugin because it has the same name as an existing coreplugin :|")
+                throw IOException()
```
rushiiMachine:
```suggestion
                Utils.showToast("External plugins are not able to override built-in coreplugins!")
                return
```

omaawr: but wait `throw IOException()` still works why remove it, it doesn't crash aliucord entirely...

rushiiMachine: Why would you even throw at all here? It's an explicitly managed error, there's no reason for it to be logged by a crash handler

rushiiMachine: Actually this error shouldn't really even be handled here but PluginFile should've never existed at all.

Just add a proper exception message to the IOException

omaawr: > Just add a proper exception message to the IOException

hmmm okay
i only really had to use ioexception because return wasn't allowed..

PR #599, opened 2025-09-03
feat: add support for avif

omardotdev:
i made it so when uploading an avif file it actually renders instead of it not being rendered


Juby210 (approved):

rushiiMachine:
There wasn't really a need to implement this as a smali patch but whatever

omaawr:
> There wasn't really a need to implement this as a smali patch but whatever

true.. but its just a list so why make an entire coreplugin for it idk

rushiiMachine:
changes to this list will require reinstalling

omaawr:
> changes to this list will require reinstalling

oh.. makes sense

PR #600, opened 2025-09-03
feat: add number of updates in the updater button

omardotdev:
sorry if my code looks bad.. i don't really know much at java


omaawr:
actually wait nop this is not okay

PR #606, opened 2025-09-16
feat: add partial support for heic

omardotdev:
for some reason it wont preview unless if you tap on it but i think thats okay.. i guess


rushiiMachine:
HEIC support is not natively supported by Discord, it shouldn't be core.

omaawr:
> HEIC support is not natively supported by Discord, it shouldn't be core.

yeah.. thats why i closed it

PR #608, opened 2025-09-21
refactor: convert `Plugin` api to Kotlin

rushiiMachine:
This is a prerequisite to deprecating a field later


Aliucord/src/main/java/com/aliucord/coreplugins/badges/SupporterBadges.kt
```diff
@@ -106,8 +106,7 @@ internal class SupporterBadges : CorePlugin(MANIFEST) {
     override fun stop(context: Context) {}
 
     private companion object {
-        val MANIFEST = Manifest().apply {
-            name = "SupporterBadges"
+        val MANIFEST = Manifest("SupporterBadges").apply {
             description = "Show badges in the profiles of contributors and donors ♡"
         }
```
zt64: I think it would be nicer to also allow setting description in the manifest constructor

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
```diff
@@ -0,0 +1,252 @@
...
+        @Suppress("UNCHECKED_CAST")
+        constructor(settings: Class<*>, type: Type) {
+            this.type = type
+            when (type) {
+                Type.PAGE ->
+                    page = settings as Class<out AppFragment>
+
+                Type.BOTTOM_SHEET ->
+                    bottomSheet = settings as Class<AppBottomSheet>
+            }
```
zt64, line 136:
```kt
            when (type) {
                Type.PAGE -> {
                    page = settings as Class<out AppFragment>
                }
                Type.BOTTOM_SHEET -> {
                    bottomSheet = settings as Class<AppBottomSheet>
                }
            }
```

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
```diff
@@ -0,0 +1,252 @@
...
+
+    /** Plugin SettingsTab */
+    class SettingsTab {
+        /** The type of this SettingsTab. PAGE is a dedicated page, BOTTOM_SHEET is a popup at the bottom of the screen. */
+        enum class Type {
+            PAGE, BOTTOM_SHEET
+        }
+
+        interface SettingsPage {
+            fun onViewBound(view: View?)
```
zt64: Wouldn't `view` be non null? Since the function is called when the view is bound, I don't think it would ever be null

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
```diff
@@ -0,0 +1,252 @@
...
+    @JvmField
+    val logger: Logger
+
+    /** SettingsTab associated with this plugin. Set this to register a settings page */
+    @JvmField
+    var settingsTab: SettingsTab? = null
+
+    /** The resources of your plugin. You need to set [.needsResources] to true to use this */
+    @JvmField
+    var resources: Resources? = null
```
zt64, line 225: I've never used resources, but shouldn't this be a internal setter? assuming it's set by Aliucord only

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
```diff
@@ -0,0 +1,252 @@
...
+    open fun start(context: Context) {
+    }
+
+    /**
+     * Called when your Plugin is stopped
+     * @param context Context
+     */
+    @Throws(Throwable::class)
+    open fun stop(context: Context) {
+    }
```
zt64, line 205: `@param context Context` I think you forgot to finish this documentation

Aliucord/src/main/java/com/aliucord/coreplugins/badges/SupporterBadges.kt
wingio: seconded

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
```diff
@@ -0,0 +1,252 @@
...
+         */
+        @JvmName("getAnnotation")
+        get() = this.javaClass.getAnnotation(AliucordPlugin::class.java)
+
+    /**
+     * Called when your Plugin is loaded
+     * @param context Context
+     */
+    @Throws(Throwable::class)
+    open fun load(context: Context) {
```
wingio, line 179: Can't all these lifecycle functions be abstract?

rushiiMachine: That would require all implementors of `Plugin` to override `load` and `unload`. I think it might also break binary compatibility

Aliucord/src/main/java/com/aliucord/entities/Plugin.kt
rushiiMachine: Unfortuantely I can't do this because of backwards compatibilty, that would make it not a field.

PR #610, opened 2025-09-23
feat(PluginDownloader): Downloading plugins from link context menu

Archimedes9500:
Finally.
Hold the message ~~(NOT LINK)~~ and press install

(once #593 gets merged I will make changes in the comments (and more if possible) to try and integrate it)


Archimedes9500:
I think this is ready now
The only issue(?) is that it doesn't work with the cursed hyperlinks that yuta makes, but I think that's a feature

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/ExtField.kt
rushiiMachine, line 1: Replace this with the accessField utils that lava added

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginDownloader.kt
rushiiMachine, line 1: Fix kotlinc warnings please

rushiiMachine:
Superseded by #679 since this PR has too many design issues and it was easier to just do it myself

PR #617, opened 2025-10-01
fix(OpenLinksExternallyFix): always open links in a separate window

Halkiion:
The purpose of this pull request is to fix an issue which occurs when opening certain app links on some devices such as YouTube where the app will open within Discord's window instead of its own separate window. This can also cause some UX issues when trying to leave the app.

This change will cause links to consistently open externally from Discord, with the exception of links intended for custom tabs.

##
Tested on: Motorola Edge 30 Ultra, Android 13


Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Bundle
+import com.aliucord.entities.CorePlugin
+import com.aliucord.patcher.before
+import de.robv.android.xposed.XC_MethodHook.MethodHookParam
+
+internal class OpenLinksExternallyFix : CorePlugin(Manifest("OpenLinksExternallyFix")) {
+    init {
+        manifest.description = "Forces app links to always open in a separate window, except for custom tabs"
```
rushiiMachine:
```suggestion
        manifest.description = "Forces app links to always open in a separate window when Chrome Custom Tabs are enabled"
```

Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
```diff
@@ -0,0 +1,33 @@
...
+            val handlers = pm.queryIntentActivities(intent, PackageManager.MATCH_DEFAULT_ONLY)
+            if (handlers.isEmpty()) {
+                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            }
+        }
+
+        patcher.before<Activity>("startActivity", Intent::class.java) { param -> handleIntent(param) }
+        patcher.before<Activity>("startActivityForResult", Intent::class.java, Int::class.javaPrimitiveType!!, Bundle::class.java) { param -> handleIntent(param) }
+        patcher.before<Context>("startActivity", Intent::class.java) { param -> handleIntent(param) }
+        patcher.before<Context>("startActivity", Intent::class.java, Bundle::class.java) { param -> handleIntent(param) }
```
rushiiMachine, line 33:
```suggestion
        patcher.before<Activity>("startActivity", Intent::class.java, ::handleIntent)
        patcher.before<Activity>("startActivityForResult", Intent::class.java, Int::class.javaPrimitiveType!!, Bundle::class.java, ::handleIntent)
        patcher.before<Context>("startActivity", Intent::class.java, ::handleIntent)
        patcher.before<Context>("startActivity", Intent::class.java, Bundle::class.java, ::handleIntent)
```
I'm fairly certain this should work, not sure though.

Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Bundle
+import com.aliucord.entities.CorePlugin
+import com.aliucord.patcher.before
+import de.robv.android.xposed.XC_MethodHook.MethodHookParam
+
+internal class OpenLinksExternallyFix : CorePlugin(Manifest("OpenLinksExternallyFix")) {
```
rushiiMachine, line 12:
```suggestion
internal class OpenLinksExternallyFix : CorePlugin(Manifest("OpenLinksExternallyFix")) {
    override val isHidden = true
```

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
Halkiion: This is not fully accurate to what the code does. It's intended for native apps which will open within Discord's task instead of starting in a new one. Links opening in custom tabs will remain unaffected.

Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
Halkiion: I would prefer to keep it the way it is as I think this suggestion requires having handleIntent as an extension. though I could take this approach if it will be cleaner and easier to maintain

Aliucord/src/main/java/com/aliucord/coreplugins/OpenLinksExternallyFix.kt
rushiiMachine: maybe mentioning that this only is an issue on certain ROMs would be good?

PR #620, opened 2025-10-04
feat(RemoveBilling): dont remove billing if user has nitro

omardotdev:
some functionalities still work for nitro users, like gifting, unsubscribing and cancelling their subscriptions, so why not enable billing for these users..

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
@@ -22,7 +22,7 @@ internal class RemoveBilling : CorePlugin(Manifest("RemoveBilling")) {
     override val isHidden = true
 
     init {
-        manifest.description = "Removes all references to billing and nitro (doesn't remove when the user has nitro)"
+        manifest.description = "Removes references to billing/nitro when user doesn't have nitro"
     }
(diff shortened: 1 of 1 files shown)


Juby210:
Why patch and then unpatch? Just don't patch if user has nitro.

omaawr:
> Why patch and then unpatch? Just don't patch if user has nitro.

oh right i can just do `return` i think

Juby210 (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -3,31 +3,36 @@ package com.aliucord.coreplugins
...
+import com.discord.widgets.stickers.UnsendableStickerPremiumUpsellDialog
+import com.lytefast.flexinput.fragment.FlexInputFragment
+import com.lytefast.flexinput.fragment.`FlexInputFragment$d`
 
 internal class RemoveBilling : CorePlugin(Manifest("RemoveBilling")) {
     override val isHidden = true
 
     init {
-        manifest.description = "Removes all references to billing and nitro"
+        manifest.description = "Removes all references to billing and nitro (doesn't remove when the user has nitro)"
```
```suggestion
        manifest.description = "Removes references to billing/nitro when user doesn't have nitro"
```

Issue #444, opened 2024-04-18
No Active Developer Badge


adamperkowski:
### Discord Account

xx0a_q

### What is it that you'd like to see?

There is no [Active Developer Profile Badge](https://support-dev.discord.com/hc/en-us/articles/10113997751447-Active-Developer-Badge) in Aliucord. 
It would be nice if it was added.

lajczi:
fixed via https://github.com/Aliucord/Aliucord/commit/021773e94ae0fc393a6a2c04eab523dd564cc5e9

PR #624, opened 2025-10-05
feat: add all new discord badges automatically

nyxiereal:
This coreplugin automatically adds all new badges to profiles, excluding all the already implemented ones.  
Fixes issue #444.

image  
image

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
@@ -17,11 +17,29 @@ internal class DiscordBadges : CorePlugin(Manifest("DiscordBadges")) {
     private val f_recyclerAdapterData by lazyField<SimpleRecyclerAdapter<*, *>>("data")
     private val f_badgeViewHolderBinding by lazyField<UserProfileHeaderView.BadgeViewHolder>("binding")
 
-    private val excludedBadgePrefixes = arrayOf(
-        "guild_booster",
-        "hypesquad",
-        "premium",
-        "bug_hunter",
+    private val excludedBadgePrefixes = setOf(
+        "guild_booster_lvl1",
+        "guild_booster_lvl2",
+        "guild_booster_lvl3",
+        "guild_booster_lvl4",
+        "guild_booster_lvl5",
+        "guild_booster_lvl6",
+        "guild_booster_lvl7",
+        "guild_booster_lvl8",
+        "guild_booster_lvl9",
+        "hypesquad_house_3",
+        "hypesquad_house_2",
+        "hypesquad_house_1",
+        "premium_tenure_72_month_v2",
+        "premium_tenure_60_month_v2",
+        "premium_tenure_36_month_v2",
+        "premium_tenure_24_month_v2",
+        "premium_tenure_12_month_v2",
+        "premium_tenure_6_month_v2",
+        "premium_tenure_3_month_v2",
+        "premium_tenure_1_month_v2",
+        "bug_hunter_level_1",
+        "bug_hunter_level_2",
         "verified_developer",
         "staff",
         "early_supporter",
@@ -40,7 +58,7 @@ internal class DiscordBadges : CorePlugin(Manifest("DiscordBadges")) {
 
             // Exclude badges that are already in aliucord
             val discordBadges = badges
-                .filter { badge -> excludedBadgePrefixes.none { badge.id.startsWith(it) } }
+                .filter { badge -> badge.id !in excludedBadgePrefixes }
                 .map { badgeData ->
                     val iconUrl = "https://cdn.discordapp.com/badge-icons/${badgeData.icon}.png"
                     Badge(0, null, badgeData.description, false, iconUrl)


Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,74 @@
...
+            val excludedBadgeIds = setOf(
+                "guild_booster",
+                "hypesquad",
+                "premium",
+                "bug_hunter",
+                "verified_developer",
+                "staff",
+                "early_supporter",
+                "partner"
+            )
```
wingio: This can be moved outside of the patch as it never changes

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,74 @@
...
+
+            // Exclude badges that are already in aliucord
+            val discordBadges = badges
+                .filter { badge -> excludedBadgeIds.none { excluded -> badge.id.contains(excluded) } }
```
wingio: could probably do `filterNot { badge -> badge.id in excludedBadgeIds }` maybe

wingio (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
nyxiereal: What do you mean by "moved outside of the patch"?

wingio: it could be declared as an instance variable alongside the cached fields

nyxiereal: i did exactly that and it just didn't exclude the badges.

nyxiereal: closing as the changes can't be done

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,74 @@
...
+    // Cached fields
+    private val f_badgesAdapter by lazyField<UserProfileHeaderView>("badgesAdapter")
+    private val f_recyclerAdapterData by lazyField<SimpleRecyclerAdapter<*, *>>("data")
+    private val f_badgeViewHolderBinding by lazyField<UserProfileHeaderView.BadgeViewHolder>("binding")
+
+    @Suppress("UNCHECKED_CAST")
+    override fun start(context: Context) {
+        // Add profile badges to the RecyclerView
+        patcher.after<UserProfileHeaderView>("updateViewState", UserProfileHeaderViewModel.ViewState.Loaded::class.java)
+        { (_, state: UserProfileHeaderViewModel.ViewState.Loaded) ->
```
rushiiMachine:
```suggestion
        patcher.after<UserProfileHeaderView>(
        	"updateViewState", 
        	UserProfileHeaderViewModel.ViewState.Loaded::class.java,
        ) { (_, state: UserProfileHeaderViewModel.ViewState.Loaded) ->
```

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
rushiiMachine: The changes can be done, move the set to an instance field of the plugin class

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,74 @@
...
+                }
+
+            val adapter = f_badgesAdapter[this] as SimpleRecyclerAdapter<Badge, UserProfileHeaderView.BadgeViewHolder>
+            val data = f_recyclerAdapterData[adapter] as MutableList<Badge>
+            data.addAll(discordBadges)
+        }
+
+        // Set image url for badge
+        patcher.after<UserProfileHeaderView.BadgeViewHolder>("bind", Badge::class.java)
+        { (_, badge: Badge) ->
```
rushiiMachine:
```suggestion
        patcher.after<UserProfileHeaderView.BadgeViewHolder>("bind", Badge::class.java) { (_, badge: Badge) ->
```

nyxiereal:
when i move the `excludedBadgeIds` next to the cached fields the badge filtering just doesn't work
image
image

it's possible that i'm just doing something wrong

rushiiMachine:
what on earth

rushiiMachine:
idrc actually, you can just leave that as-is then. Not worth the effort debugging.
Use an array instead of a set though, smaller overhead

Aliucord/src/main/java/com/aliucord/api/rn/user/ProfileBadge.kt
```diff
@@ -0,0 +1,9 @@
+package com.aliucord.api.rn.user
+
+@Suppress("unused")
```
zt64: Shouldn't be needed since it's used in the code

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.aliucord.entities.CorePlugin
+import com.aliucord.patcher.*
+import com.aliucord.utils.lazyField
+import com.discord.databinding.UserProfileHeaderBadgeBinding
+import com.discord.utilities.views.SimpleRecyclerAdapter
+import com.discord.widgets.user.Badge
+import com.discord.widgets.user.profile.UserProfileHeaderView
+import com.discord.widgets.user.profile.UserProfileHeaderViewModel
+
+internal class DiscordBadges : CorePlugin(MANIFEST) {
```
zt64: The manifest should just be defined here inline without the variable. It's a little uglier but eliminates an unnecessary companion object

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,78 @@
...
+    // Cached fields
+    private val f_badgesAdapter by lazyField<UserProfileHeaderView>("badgesAdapter")
+    private val f_recyclerAdapterData by lazyField<SimpleRecyclerAdapter<*, *>>("data")
+    private val f_badgeViewHolderBinding by lazyField<UserProfileHeaderView.BadgeViewHolder>("binding")
```
zt64, line 18: Avoid using snake_case naming like this. It's more coherent to just use the longer name, e.g. `badgesAdapterField`

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,78 @@
...
+        patcher.after<UserProfileHeaderView.BadgeViewHolder>("bind", Badge::class.java)
+        { (_, badge: Badge) ->
+            val url = badge.objectType
+
+            // Check that badge is ours (has icon = 0 and url set)
+            if (badge.icon != 0 || url == null) return@after
+
+            val binding = f_badgeViewHolderBinding[this] as UserProfileHeaderBadgeBinding
+            val imageView = binding.b
+            imageView.setCacheableImage(url)
```
zt64: No need for a variable here, just use `binding.b` directly. Add a comment stating what `binding.b` is.

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,78 @@
...
+            "updateViewState", 
+            UserProfileHeaderViewModel.ViewState.Loaded::class.java,
+        ) { (_, state: UserProfileHeaderViewModel.ViewState.Loaded) ->
+            val profile = state.userProfile as? RNUserProfile ?: return@after
+            val badges = profile.badges ?: return@after
+
+            // Exclude badges that are already in aliucord
+            val discordBadges = badges
+                .filterNot { badgeData -> 
+                    excludedBadgeIds.any { excludedId -> badgeData.id.contains(excludedId) }
```
zt64: You can use `in` as an alternative to the longer `contains`:
```kt
excludedId in badgeData.id
```

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.discord.widgets.user.profile.UserProfileHeaderView
+import com.discord.widgets.user.profile.UserProfileHeaderViewModel
+
+internal class DiscordBadges : CorePlugin(MANIFEST) {
+    // Cached fields
+    private val f_badgesAdapter by lazyField<UserProfileHeaderView>("badgesAdapter")
+    private val f_recyclerAdapterData by lazyField<SimpleRecyclerAdapter<*, *>>("data")
+    private val f_badgeViewHolderBinding by lazyField<UserProfileHeaderView.BadgeViewHolder>("binding")
+
+    private val excludedBadgeIds = arrayOf(
```
zt64: Would a List work here?

nyxiereal: rushii told me to use an array instead of a set :sob:

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.discord.utilities.views.SimpleRecyclerAdapter
+import com.discord.widgets.user.Badge
+import com.discord.widgets.user.profile.UserProfileHeaderView
+import com.discord.widgets.user.profile.UserProfileHeaderViewModel
+
+internal class DiscordBadges : CorePlugin(Manifest("DiscordBadges")) {
+    // Cached fields
+    private val badgesAdapter by lazyField<UserProfileHeaderView>("badgesAdapter")
+    private val recyclerAdapterData by lazyField<SimpleRecyclerAdapter<*, *>>("data")
+    private val badgeViewHolderBinding by lazyField<UserProfileHeaderView.BadgeViewHolder>("binding")
```
rushiiMachine: I know what zt said but this is even worse, it doesn't convey the fact these are fields at all.
Either go back to `f_` (which is bit of a personal convention I have) or add a `Field` suffix to the end.

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
```diff
@@ -0,0 +1,70 @@
...
+            UserProfileHeaderViewModel.ViewState.Loaded::class.java,
+        ) { (_, state: UserProfileHeaderViewModel.ViewState.Loaded) ->
+            val profile = state.userProfile as? RNUserProfile ?: return@after
+            val badges = profile.badges ?: return@after
+
+            // Exclude badges that are already in aliucord
+            val discordBadges = badges
+                .filterNot { badgeData -> 
+                    excludedBadgeIds.any { excludedId -> excludedId in badgeData.id }
+                }
```
rushiiMachine:
```suggestion
                .filter { badgeData -> badgeData.id !in excludedBadgeIds }
```

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
zt64: I believe this was resolved by mistake, as it's still present.

nyxiereal: Oops

Aliucord/src/main/java/com/aliucord/coreplugins/badges/DiscordBadges.kt
nyxiereal: This will not work, unless we include every single badge prefix in the `excludedBadgePrefixes` array, which could negatively impact performance.

nyxiereal: I can do that if you want me to, it would be a cleaner solution when I think about it.

rushiiMachine: Yea i forgot to delete this review, I implemented a better version in this commit [`e64c1af` (#624)](https://github.com/Aliucord/Aliucord/pull/624/commits/e64c1af50d4241748b7a0ec19ff055af3db7e99b)

PR #626, opened 2025-10-15
feat(RemoveBilling): remove more upsells & cleanup

omardotdev:
this patches the get nitro button when holding down emojis aswell as the "learn about nitro" button when uploading a huge file which is more than 50mb, this also uses the new viewutils extension and uses `textview.visibility = View.GONE` instead of `textview.setVisibility(View.GONE)`

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
@@ -98,7 +98,7 @@ internal class RemoveBilling : CorePlugin(Manifest("RemoveBilling")) {
 
         // Remove "Get Nitro" button when holding down on emojis
         val getBinding = WidgetEmojiSheet::class.java.getDeclaredMethod("getBinding")
-        getBinding.isAccessible = true
+            .apply { isAccessible = true }
 
         patcher.after<WidgetEmojiSheet>("configureButtons", Boolean::class.java, Boolean::class.java, Guild::class.java) { param ->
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
rushiiMachine, line 1: - Don't inline multiple statements at once
- When possible, use the .apply {} lambda to avoid defining extra variables that won't be used elsewhere
```kt
layout.findViewById<...>("...").apply { ... }
// or
layout.findViewById<...>("...")
    .apply { ... }
// or
layout.findViewById<...>("...").apply {
    ...
}
```

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/RemoveBilling.kt
```diff
@@ -52,37 +57,54 @@ internal class RemoveBilling : CorePlugin(Manifest("RemoveBilling")) {
...
+
+        // Remove the "Learn about Nitro" button when a huge file is uploaded
+        patcher.after<ImageUploadFailedView>("onViewBound", View::class.java) { (_, view: View) ->
+            view.findViewById<RelativeLayout>("image_upload_failed_nitro_wrapper")
+                .apply { visibility = View.GONE }
+        }
+
+        // Remove "Get Nitro" button when holding down on emojis
+        val getBinding = WidgetEmojiSheet::class.java.getDeclaredMethod("getBinding")
+        getBinding.isAccessible = true
```
```suggestion
        val getBinding = WidgetEmojiSheet::class.java.getDeclaredMethod("getBinding")
            .apply { isAccessible = true }
```

PR #632, opened 2025-10-24
feat: add missing experiments

CatPieLeaf:
Added some missing experiment defaults

Enabled `2021-10_premium_guild_member_profiles` to allow editing premium guild profiles' descriptions.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
@@ -18,8 +18,11 @@ internal class ExperimentDefaults : CorePlugin(Manifest("ExperimentDefaults")) {
         // Thanks Dolfies for helping with this
         // Full list of v126.21 experiments: https://discord.com/channels/811255666990907402/811262084968742932/1276337938661118062
         val newOverrides = arrayOf(
+            "2020-01_mobile_invite_suggestion_compact" to 0,
             "2020-09_threads" to 1,
+            "2020-12_android_guild_channel_invite_sheet" to 0,
             "2020-12_guild_delete_feedback" to 0, // Analytics are not sent anyway, this is useless
+            "2020-12_invite_to_gdm" to 0,
             "2021-02_view_threads" to 1,
             "2021-03_android_extend_invite_expiration" to 1, // This matches with desktop
             "2021-03_android_media_sink_wants" to 1,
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
```diff
@@ -18,21 +18,28 @@ internal class ExperimentDefaults : CorePlugin(Manifest("ExperimentDefaults")) {
...
+            "2021-06_preview_promotions" to 1,
             "2021-06_reg_bailout_to_email_android" to 1, // Doesn't really matter
+            "2021-07_network_action_logging_android" to 0,
+            "2021-07_threads_only_channel" to 1,
             "2021-08_hub_multi_domain_mobile" to 1,
             "2021-08_threads_permissions" to 1,
             "2021-09_android_app_commands_frecency" to 1,
             "2021-09_android_sms_autofill" to 1,
             "2021-10_android_attachment_bottom_sheet" to 1,
+            "2021-10_premium_guild_member_profiles" to 1,
```
rushiiMachine, line 42: I don't think this is correct? Don't we have a more up to date plugin that enables proper editing of profiles

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
```diff
@@ -18,21 +18,28 @@ internal class ExperimentDefaults : CorePlugin(Manifest("ExperimentDefaults")) {
...
             "2020-12_guild_delete_feedback" to 0, // Analytics are not sent anyway, this is useless
+            "2020-12_invite_to_gdm" to 1,
             "2021-02_view_threads" to 1,
             "2021-03_android_extend_invite_expiration" to 1, // This matches with desktop
             "2021-03_android_media_sink_wants" to 1,
             "2021-03_stop_offscreen_video_streams" to 1,
             "2021-04_contact_sync_android_main" to 1, // This system is deprecated I think but still functional?
             "2021-06_desktop_school_hubs" to 1,
             "2021-06_hub_email_connection" to 1,
+            "2021-06_preview_promotions" to 1,
```
rushiiMachine, line 33: What does this do?

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
```diff
@@ -18,21 +18,28 @@ internal class ExperimentDefaults : CorePlugin(Manifest("ExperimentDefaults")) {
         // Thanks Dolfies for helping with this
         // Full list of v126.21 experiments: https://discord.com/channels/811255666990907402/811262084968742932/1276337938661118062
         val newOverrides = arrayOf(
+            "2020-01_mobile_invite_suggestion_compact" to 1,
             "2020-09_threads" to 1,
+            "2020-12_android_guild_channel_invite_sheet" to 1,
             "2020-12_guild_delete_feedback" to 0, // Analytics are not sent anyway, this is useless
+            "2020-12_invite_to_gdm" to 1,
```
rushiiMachine: When I was first making this list of experiments, these were specifically omitted cause people didn't want it.

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
CatPieLeaf: Preview endpoint for discord nitro rewards (free random codes and gifts that you get with nitro). probably useless since billing is disabled

Or not, since nitro gifts menu is still present on aliucord

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
CatPieLeaf: I only added to match RN discord. That's the updated behavior anyway. They can disable if they don't like it

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
CatPieLeaf: Which plugin you refer to? With the plugins i know, editing descriptions on premium profiles isn't possible, only username and profile picture

Aliucord/src/main/java/com/aliucord/coreplugins/ExperimentDefaults.kt
rushiiMachine: This should be removed

CatPieLeaf: Okey, i'll set it to 0 just to keep all the missing experiments.
Tell me if you prefer the lines actually removed

PR #637, opened 2025-10-28
fix: migrate from rawgithub to jsdelivr and builds.aliucord.com to avoid rate limits

CatPieLeaf:
Migrates all raw.githubusercontent.com urls to cdn.jsdelivr.net and builds.aliucord.com, to solve all `429: Too Many Requests` errors on CGNAT networks after recent GitHub rate limit enforcements.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
+++ b/Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.java
@@ -118,9 +118,14 @@ public class PluginUpdater {
     public static UpdateInfo getUpdateInfo(Plugin plugin) throws Exception {
         Plugin.Manifest manifest = plugin.getManifest();
         if (manifest.updateUrl == null || manifest.updateUrl.equals("")) return null;
+
+        String updateUrl = manifest.updateUrl.replaceFirst(
+            "https://raw\\.githubusercontent\\.com/(.+?)/(.+?)/(.+)",
+            "https://cdn.jsdelivr.net/gh/$1/$2@$3"
+        );
         String name = plugin.getName();
 
-        CachedData cached = cache.get(manifest.updateUrl);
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -209,7 +209,9 @@ private String getGithubUrl(Plugin plugin) {
             return plugin
                 .getManifest().updateUrl
                 .replace("raw.githubusercontent.com", "github.com")
-                .replaceFirst("/builds.*", "");
+                .replace("cdn.jsdelivr.net/gh/", "github.com/")
+                .replaceFirst("/builds.*", "")
+                .replaceFirst("@.*", "");
```
rushiiMachine: Can you use a regex instead of a flimsy replace like this

installer/lib/github.dart
```diff
@@ -140,5 +140,5 @@ class GithubAPI with ChangeNotifier {
   }
 
   String getDownloadUrl(String ref, String file) =>
-      'https://raw.githubusercontent.com/$_org/$_repo/$ref/$file';
+      'https://cdn.jsdelivr.net/gh/$_org/$_repo@$ref/$file';
```
rushiiMachine: Undo this, legacy installer is not used anymore.

rushiiMachine (changes requested):
todo for me later:
- [ ] bump core and injector version

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
CatPieLeaf: 👍

installer/lib/github.dart
CatPieLeaf: 👍

PR #638, opened 2025-10-29
feat(Decorations): implement avatar decorations

cillynder:
Adds avatar decorations to the message author, members list, DMs list, and the profile header (animated). All decorations will be individually toggleable via the plugin settings.

Also added some padding utils, and an abstract API for other decorations to come.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/decorations/Decorator.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/decorations/Decorator.kt
@@ -12,7 +12,7 @@ import com.discord.widgets.chat.list.entries.MessageEntry
 import com.discord.widgets.user.profile.UserProfileHeaderView
 import com.discord.widgets.user.profile.UserProfileHeaderViewModel
 
-abstract class Decorator {
+internal abstract class Decorator {
     open fun patch(patcher: PatcherAPI) {}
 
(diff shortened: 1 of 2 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/decorations/Decorator.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.discord.widgets.channels.list.items.ChannelListItemPrivate
+import com.discord.widgets.channels.memberlist.adapter.ChannelMembersListAdapter
+import com.discord.widgets.channels.memberlist.adapter.ChannelMembersListViewHolderMember
+import com.discord.widgets.chat.list.adapter.WidgetChatListAdapter
+import com.discord.widgets.chat.list.adapter.WidgetChatListAdapterItemMessage
+import com.discord.widgets.chat.list.entries.MessageEntry
+import com.discord.widgets.user.profile.UserProfileHeaderView
+import com.discord.widgets.user.profile.UserProfileHeaderViewModel
+
+abstract class Decorator {
```
rushiiMachine: internal?

cillynder: oops

rushiiMachine (approved):

PR #639, opened 2025-10-30
fix(Polls): fix permission checking when user is owner/admin

cillynder:
#625 didn't probably handle cases where the user is the owner. This is because in such cases, Discord uses a `Permission.ALL` value that does not include the polls permission. Unfortunately as it's a static final value, it gets inlined, so a smali patch would be the easiest way to go about it.

The new permission includes the SEND_POLLS permission and also the new PIN_MESSAGES permission.

I have included a check to make sure the user has the smali patches necessary to support the permission check. This is optional and can be removed if it's better to do so.

Also, accessor comments are now disabled during the baksmali process. In Manager, these comments weren't generated, so the source files will be different and lead to patching failures in some cases (for instance, the patch for `WidgetChannelSettingsEditPermissionsModel$Companion$getForRole$1` in this PR will fail without this change).

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/Polls.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/Polls.kt
@@ -346,6 +346,10 @@ internal class Polls : CorePlugin(Manifest("Polls")) {
             val ctx = flexInputFragment.requireContext()
             val pages = flexInputFragment.r.toMutableList()
 
+            // This checks a user's permissions before adding the polls tab
+            // Without 1.3.1 smali patches, these permission checks will fail to account for an
+            // admin/owner user, so we disable them for older versions.
+            // ref: https://github.com/Aliucord/Aliucord/pull/639
             if (ManagerBuild.hasPatches("1.3.1")) {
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/Polls.kt
```diff
@@ -346,19 +346,20 @@ internal class Polls : CorePlugin(Manifest("Polls")) {
...
-            val channel = StoreStream.getChannelsSelected().selectedChannel
-                ?: return@after
-            val isPrivate = ChannelUtils.B(channel)
-            val isSystem = ChannelUtils.E(channel)
-            val permissions = StoreStream.getPermissions().permissionsByChannel[channel.id]
-            val hasPermission = PermissionUtils.can(SEND_POLLS_PERMISSION, permissions)
-            val permitted = (isPrivate && !isSystem) || hasPermission
-
-            if (!permitted) return@after
+            if (ManagerBuild.hasPatches("1.3.1")) {
```
rushiiMachine, line 353: Can you add a small comment here with a link to this pr or a small explanation of what's going on

PR #643, opened 2025-11-05
feat: safe mode

Canny1913:
Allows disabling plugins and optional core plugins at once. Current implementation doesn't allow `Plugins` page in settings to work properly.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
@@ -38,9 +38,7 @@ internal class PluginFile(val plugin: String) : File("${Constants.PLUGINS_PATH}/
                         PluginManager.stopPlugin(plugin)
                         PluginManager.unloadPlugin(plugin)
                     }
-                    if (safeMode) {
-                        Utils.showToast("Plugin $plugin is disabled due to safe mode.")
-                    } else {
+                    if (!safeMode) {
                         PluginManager.loadPlugin(Utils.appContext, this)
                         if (PluginManager.isPluginEnabled(plugin))
                             PluginManager.startPlugin(plugin)
@@ -48,7 +46,9 @@ internal class PluginFile(val plugin: String) : File("${Constants.PLUGINS_PATH}/
                             PluginManager.enablePlugin(plugin)
                     }
                     Utils.showToast("Plugin $plugin successfully ${if (isReinstall) "re" else ""}installed!")
-
+                    if (safeMode) {
+                        Utils.showToast("Plugin not loaded due to safe mode!")
+                    }
                     if (PluginManager.plugins[plugin]?.requiresRestart() == true)
                         Utils.promptRestart()
 


Aliucord/src/main/java/com/aliucord/PluginManager.kt
```diff
@@ -323,7 +337,9 @@ object PluginManager {
             UploadSize(),
         )
 
-        corePlugins.forEach { p ->
+        corePlugins.filter { p ->
+            p.isRequired && isSafeModeEnabled()
+        }.forEach { p ->
```
Sculas, line 334: That filter doesn't look correct. Also, why not return early?
```suggestion
        if (isSafeModeEnabled()) return
        corePlugins.forEach { p ->
```

Canny1913: Enforced coreplugins should bypass safe mode as they are essential to ensure app stability.
But the filter is indeed not correct :sob:

Canny1913:
Draft due to planned changes.

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -299,6 +304,25 @@ public void onViewBound(View view) {
...
+            safeModeNotice.setAllCaps(false);
+            safeModeNotice.setText("This page won't work while safe mode is on.\n Use Aliucord Manager to change plugin settings.");
+            safeModeNotice.setTypeface(ResourcesCompat.getFont(context, Constants.Fonts.whitney_semibold));
+            safeModeNotice.setGravity(Gravity.CENTER);
+
+            Button openManagerButton = new Button(context);
+            openManagerButton.setText("Open Manager");
+            openManagerButton.setOnClickListener(widget -> {
+                Intent intent = new Intent("com.aliucord.manager.OPEN_PLUGINS");
+                intent.setClassName("com.aliucord.manager", "com.aliucord.manager.MainActivity");
```
rushiiMachine: You should set the `aliucord.packageName` intent extra to the current package name here. It's unused atm but will be in the future.

Aliucord/src/main/java/com/aliucord/PluginManager.kt
```diff
@@ -323,7 +328,9 @@ object PluginManager {
             UploadSize(),
         )
 
-        corePlugins.forEach { p ->
+        corePlugins.filter { p ->
+            !isSafeModeEnabled() || p.isRequired
```
rushiiMachine: `isSafeModeEnabled()` should be reused here, as it's not a simple variable read

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
```diff
@@ -37,11 +37,13 @@ internal class PluginFile(val plugin: String) : File("${Constants.PLUGINS_PATH}/
...
-                        PluginManager.startPlugin(plugin)
-                    else
-                        PluginManager.enablePlugin(plugin)
+                    if (!PluginManager.isSafeModeEnabled()) {
+                        PluginManager.loadPlugin(Utils.appContext, this)
+                        if (PluginManager.isPluginEnabled(plugin))
+                            PluginManager.startPlugin(plugin)
+                        else
+                            PluginManager.enablePlugin(plugin)
+                    }
```
rushiiMachine, line 47: You should display a toast here if safe mode is on, "plugin downloaded but not loaded due to safe mode" or something

Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -182,6 +185,43 @@ public static void init(AppActivity activity) {
...
+            Context context = indicator.requireContext();
+
+            try {
+                Method bindingGetter = indicator.getClass().getDeclaredMethod("getBinding");
+                bindingGetter.setAccessible(true);
+                Field indicatorStateField = indicator.getClass().getDeclaredField("indicatorState");
+                indicatorStateField.setAccessible(true);
+
+                var binding = (WidgetGlobalStatusIndicatorBinding) bindingGetter.invoke(indicator);
+                var indicatorState = (WidgetGlobalStatusIndicatorState) indicatorStateField.get(indicator);
```
rushiiMachine: Use `ReflectUtils` here, it'll simplify this

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
Canny1913: That is way too long for a toast message. a Dialog could be used instead?

Canny1913:
Draft because i don't like the current implementation of Open Manager button.

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -372,4 +366,14 @@ public void onTextChanged(CharSequence s, int start, int before, int count) { }
...
             });
     }
+    public void onOpenManagerClick(View view) {
+        Intent intent = new Intent("com.aliucord.manager.OPEN_PLUGINS");
+        intent.setClassName("com.aliucord.manager", "com.aliucord.manager.MainActivity");
+        intent.putExtra("aliucord.packageName", requireContext().getPackageName());
+        try {
+            startActivity(intent);
+        } catch (ActivityNotFoundException e) {
+            Utils.showToast("Aliucord Manager is not installed.");
```
Canny1913: Could've added a Dialog which redirected users to Manager repo but bleh

Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -518,4 +558,22 @@ private static boolean checkPermissions(AppCompatActivity activity) {
         ).launch(perm);
         return false;
     }
+    private static void safeModeDialog(Fragment fragment) {
```
rushiiMachine, line 562: Missing empty line

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
rushiiMachine: I think you should do

you could use https://developer.android.com/reference/android/content/Intent#resolveActivity(android.content.pm.PackageManager) to check whether manager exists rather than catching exception if you want

Aliucord/src/main/java/com/aliucord/coreplugins/plugindownloader/PluginFile.kt
rushiiMachine: thats too much i think

maybe "Plugin disabled due to safe mode!"

rushiiMachine (approved):

PR #650, opened 2025-11-18
fix(Decorations/Avatar): fix alignment in DMs list

cillynder:
Avatar decoration in DMs list was unaligned with actual avatar

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/decorations/avatar/AvatarDecorator.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/decorations/avatar/AvatarDecorator.kt
@@ -117,8 +117,11 @@ internal class AvatarDecorator() : Decorator() {
         // Removes padding from layout to make space for deco
         layout.padding -= memberListSpacing.dp
 
+        val decoMargin: Int
+
         // Re-add left padding to avatar view, that was removed above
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/decorations/avatar/AvatarDecorator.kt
```diff
@@ -117,16 +117,19 @@ internal class AvatarDecorator() : Decorator() {
...
+            decoMargin = marginStart
             marginStart += memberListSpacing.dp
         }
 
         createDecoView(layout.context, false).addTo(layout, layout.indexOfChild(avatarView) + 1) {
             layoutParams = RelativeLayout.LayoutParams(0, 0).apply {
                 height = size
                 width = size
-                marginStart = memberListSpacing.dp
+                marginStart = decoMargin
```
rushiiMachine, line 132: Would it not be simpler to just
```suggestion
                marginStart = avatarView.layoutParams.marginStart
```
?

cillynder: Would need to cast avatarView.layoutParams for that, and also reordering it, I don't think it's worth it imo

PR #677, opened 2026-02-02
fix(TokenLogin): trim token input

Canny1913:
Fixes tokens obtained from `Token` plugin not working with TokenLogin.


Ushie:
Could also remove quotes since people typically copy them with quotes

Canny1913:
That is user fault. This just fixes an unintentional bug leading to weird side effects.

rushiiMachine (approved):

PR #681, opened 2026-02-03
feat(ComponentsV2): add cv2 api models to injector

cillynder:
Adds all the new components' api models into injector.

~~Also bumps Injector to 2.4.0~~

### DIFF
--- a/Injector/src/main/java/com/discord/api/botuikit/SelectV2DefaultValueType.kt
+++ b/Injector/src/main/java/com/discord/api/botuikit/SelectV2DefaultValueType.kt
@@ -1,0 +1,7 @@
+package com.discord.api.botuikit
+
+enum class SelectV2DefaultValueType {
+    @b.i.d.p.b("user") USER,
+    @b.i.d.p.b("role") ROLE,
+    @b.i.d.p.b("channel") CHANNEL,
+}
--- a/Injector/src/main/java/com/discord/api/botuikit/SelectV2DefaultValueType.kt
+++ b/Injector/src/main/java/com/discord/api/botuikit/SelectV2DefaultValueType.kt
@@ -1,7 +1,9 @@
 package com.discord.api.botuikit
 
+import b.i.d.p.b as SerializedName
+
 enum class SelectV2DefaultValueType {
-    @b.i.d.p.b("user") USER,
-    @b.i.d.p.b("role") ROLE,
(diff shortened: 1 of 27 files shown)


mantikafasi (approved):

rushiiMachine:
I unintentionally broke injector early init so I need to fix that first before Injector version is bumped.

Injector/src/main/java/com/discord/api/botuikit/SelectV2DefaultValueType.kt
```diff
@@ -0,0 +1,7 @@
+package com.discord.api.botuikit
+
+enum class SelectV2DefaultValueType {
+    @b.i.d.p.b("user") USER,
+    @b.i.d.p.b("role") ROLE,
+    @b.i.d.p.b("channel") CHANNEL,
```
rushiiMachine: I think importing this as `@SerializedName` or whatever it is would be a better idea

cillynder:
> I unintentionally broke injector early init so I need to fix that first before Injector version is bumped.

It'll be fine for now since this is going in a feature branch anyway

rushiiMachine:
> refactor(Polls): remove unnecessary SerializedName
> The default gson instance already converts casing

I don't particularly mind this but are you sure we should be relying on essentially undocumented behavior? It could cause future confusion if they need to be modified. I didn't even know gson did that until now

cillynder:
> > refactor(Polls): remove unnecessary SerializedName
> > The default gson instance already converts casing
> 
> I don't particularly mind this but are you sure we should be relying on essentially undocumented behavior? It could cause future confusion if they need to be modified. I didn't even know gson did that until now

It's how DiscordKt configures it, similar to what we have with `GsonUtils.gsonRestApi`. I think it'll be safe to use

rushiiMachine (approved):
Aside from the version bump

rushiiMachine:
Could I remove the version bump and merge this now? I was planning on bumping Injector to v3.0.0 once #700 is merged anyways

cillynder:
Sure! Let me just rebase the branch

Ushie:
> I unintentionally broke injector early init so I need to fix that first before Injector version is bumped.

Does this depend on anything from there? If injector is broken then it would be good to revert and move the changes from there to #700

rushiiMachine:
> Does this depend on anything from there? 

no, we could revert the injector bump in this pr for now and merge this

PR #686, opened 2026-02-07
fix: support slowmode permission

cillynder:
Only adds logic handling the permission, does not have permission management etc like new pins.


Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -396,6 +398,16 @@ public static void init(AppActivity activity) {
...
+        Patcher.addPatch(PermissionUtils.class,
+            "hasBypassSlowmodePermissions",
+            new Class[]{ Long.class, StoreSlowMode.Type.class },
+            new InsteadHook(param -> {
+                Long permissions = (Long)param.args[0];
+                return PermissionUtils.can(Permission.ADMINISTRATOR, permissions)
+                    || PermissionUtils.can(1L << 52, permissions);
+            })
+        );
+
```
rushiiMachine: A comment explanation of this patch + a constant for the slowmode permission bit would be great

Aliucord/src/main/java/com/aliucord/Main.java
rushiiMachine, line 1: you forgot unused imports

PR #691, opened 2026-02-08
refactor: reorganize core patches

zt64:
Cleans up a lot of the core patches by moving them into a dedicated CoreFixes and CoreBase (this name isn't final, if there's something better tell me) Doing this massively cleans up Main.java and reduces the number of classes for small patches.

CoreFixes.kt is intended for all the small patches that fix something
CoreBase.kt is intended for stuff that adds functionality like the Aliucord settings section
I'm not sure if I like CoreBase personally, but it's an improvement over having everything shoved in Main.java

I've been testing on my own installation and haven't had any errors but someone else should test to confirm.


rushiiMachine:
Note that each logically separate fix should be in it's own try/catch block to ensure that one failing won't cause all of the others to break as well. You could write a wrapper for that or something.

And maybe separate different fixes into their own functions as well.

Aliucord/src/main/java/com/aliucord/coreplugins/AlignThreadsIcon.kt
rushiiMachine, line 1: I think this should be merged into CoreFixes, don't really see a point in keeping this separate

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFeatures.kt
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        patcher.after<WidgetDebugging.Adapter.Item>(
+            "onConfigure",
+            Int::class.javaPrimitiveType!!,
+            LoggedItem::class.java
+        ) { (param, _: Any, loggedItem: LoggedItem) ->
+            val th = loggedItem.m ?: return@after
+
+            val logMessage = (debugItemBinding[param.thisObject] as WidgetDebuggingAdapterItemBinding).b
```
rushiiMachine: This can be cleaned up with ReflectDelegates

Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -96,34 +96,13 @@ public static void preInit(AppActivity activity) throws NoSuchMethodException {
...
-                        return;
-                    default:
-                        try {
-                            ReflectUtils.setField($this, "theme", "dark");
-                        } catch (Exception e) {
-                            logger.error("Failed to fix ModelUserSettings theme", e);
-                        }
-                }
-            }));
-        }
```
rushiiMachine, line 127: This needs to remain here, it's critical and should be patched asap.
If something further along the line breaks with PluginManager, this would not be applied and would break stock

Aliucord/src/main/java/com/aliucord/PluginManager.kt
```diff
@@ -307,28 +308,21 @@ object PluginManager {
...
-            ShowReplyMention(),
+            ShowReplyMentionFix(),
             SlowmodeFix(),
-            StickerCrashFix(),
             SupportWarn(),
             SupporterBadges(),
             TokenLogin(),
-            UploadSize(),
-            VoiceFix(),
+            UploadSize()
```
rushiiMachine: missing comma

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -0,0 +1,267 @@
...
+    private fun fixWebpEmojis() = tryPatch("Fix emoji formats") {
+        // Support webp emojis by forcing every emoji to be webp
+        patcher.patch(
+            ModelEmojiCustom::class.java,
+            "getImageUri",
+            arrayOf(Long::class.javaPrimitiveType!!, Boolean::class.javaPrimitiveType!!, Int::class.javaPrimitiveType!!),
+            InsteadHook { param ->
+                "https://cdn.discordapp.com/emojis/${param.args[0]}.webp?size=${param.args[2]}&animated=${param.args[1]}"
+            }
+        )
```
rushiiMachine: Does patcher.instead not work here? or is it because it's static or something

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    private fun fixMemberListGroups() = tryPatch("Fix member list groups") {
+        val memberListGroupsField = ChannelMemberList::class.java.getDeclaredField("groups")
+            .apply { isAccessible = true }
+        @Suppress("UNCHECKED_CAST")
+        patcher.after<ChannelMemberList>("setGroups", List::class.java, Function1::class.java) {
+            val groupsMap = memberListGroupsField[this] as Map<String, MemberListRow>
+            groupIndices.forEach { (idx, id) -> rows[idx] = groupsMap[id] }
+        }
```
rushiiMachine: Use ReflectDelegates

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -0,0 +1,267 @@
...
+            }
+        }
+    }
+
+    private fun fixStockEmojis() = tryPatch("Fix built-in emojis") {
+        // Patch to repair built-in emotes is needed because installer doesn't recompile resources,
+        // so they stay in package com.discord instead of apk package name
+        Patcher.addPatch(ModelEmojiUnicode::class.java, "getImageUri", arrayOf(String::class.java, Context::class.java),
+            InsteadHook { param -> "res:///${getResId("emoji_${param.args[0]}", "raw")}" }
+        )
```
rushiiMachine: use kotlin patcher api

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -0,0 +1,267 @@
...
+
+    private fun fixKeyboardCrash() {
+        // not sure why this happens, reported on Android 15 Beta 4
+        // java.lang.IllegalArgumentException: Animators cannot have negative duration: -1
+        //   at android.view.ViewPropertyAnimator.setDuration(ViewPropertyAnimator.java:266)
+        //   at com.discord.widgets.chat.input.SmoothKeyboardReactionHelper$Callback.onStart(SmoothKeyboardReactionHelper.kt:5)
+        //   at android.view.View.dispatchWindowInsetsAnimationStart(View.java:12671)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+            tryPatch("Fix keyboard animation crash on Android 15") {
+                Patcher.addPatch(
```
rushiiMachine: use kotlin patcher api

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
zt64: I tried but it doesn't work for these functions

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
zt64: I tried but it doesn't work for these functions

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
rushiiMachine: you can still use patcher.patch(member, hook) in that case then

Canny1913: > I tried but it doesn't work for these functions

it doesnt work because its a static method
as a workaround you can pass a nullable version of class type to patcher.instead

rushiiMachine: do this @zt64

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -199,7 +198,7 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
         patcher.after<`WidgetChannelListModel$Companion$guildListBuilder$$inlined$forEach$lambda$1$1`>("invoke") { param ->
             val builder = this.`this$0`
             val threadId = this.`$textChannel`.id
-            val mentionCount = builder.`$mentionCounts$inlined`[threadId] as Int? ?: return@after
+            val mentionCount = builder.`$mentionCounts$inlined`[threadId] as? Int? ?: return@after
```
rushiiMachine, line 202: curious why'd you do this, i don't see how it would be any other type other than Int or null

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -126,15 +103,14 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
     }
 
     private fun fixWebpEmojis() = tryPatch("Fix emoji formats") {
+        ModelEmojiCustom::getImageUri
```
rushiiMachine: this seems like a mistake

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -126,15 +103,14 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+        patcher.instead<ModelEmojiCustom?>(
             "getImageUri",
-            arrayOf(Long::class.javaPrimitiveType!!, Boolean::class.javaPrimitiveType!!, Int::class.javaPrimitiveType!!),
-            InsteadHook { param ->
-                "https://cdn.discordapp.com/emojis/${param.args[0]}.webp?size=${param.args[2]}&animated=${param.args[1]}"
-            }
-        )
+            Long::class.javaPrimitiveType!!, Boolean::class.javaPrimitiveType!!, Int::class.javaPrimitiveType!!,
+        ) { param ->
+            "https://cdn.discordapp.com/emojis/${param.args[0]}.webp?size=${param.args[2]}&animated=${param.args[1]}"
```
rushiiMachine: unwrap args with patcher extensions?

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
zt64: Android Studio was yelling about it, other option is just suppressing the warning

rushiiMachine: what was the warning

zt64: "Replace the unsafe cast with the safe one". Using `as?` saves one line since it's already nullable instead of adding a suppress annotation

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFeatures.kt
```diff
@@ -0,0 +1,276 @@
...
+            "configureUI",
+            WidgetGlobalStatusIndicatorViewModel.ViewState::class.java,
+        ) { param ->
+            if (!isSafeModeEnabled()) return@before
+
+            val indicator = param.thisObject as WidgetGlobalStatusIndicator
+            val context = indicator.requireContext()
+
+            val binding = ReflectUtils.invokeMethod(indicator, "getBinding") as WidgetGlobalStatusIndicatorBinding
+            val indicatorState = ReflectUtils.getField(indicator, "indicatorState") as WidgetGlobalStatusIndicatorState
```
cillynder: These could be converted into delegates

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -0,0 +1,281 @@
...
+            String::class.java,
+            Int::class.javaPrimitiveType!!,
+            String::class.java,
+        ) { (param, _: String, _: Int, mode: String) ->
+            if (mode == "xsalsa20_poly1305") {
+                param.args[2] = "xsalsa20_poly1305_lite_rtpsize"
+            }
+        }
+    }
+
```
cillynder, line 271: This could be removed since it doesn't help with anything after DAVE enforcement. It will eventually be inside the coreplugin enabling support for it.

cillynder (changes requested):
`findViewById(getResId("...", "id"))` can be replaced by the shorthand `findViewById("...")` in `ViewUtils.findViewById`, which also casts it properly and avoids platform types.

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
rushiiMachine: Maybe do this in a separate commit to document the removal?

cillynder: If so, might as well just keep it then and remove it in the eventual voice PR.

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -109,13 +125,12 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
     private fun fixWebpEmojis() = tryPatch("Fix emoji formats") {
-        ModelEmojiCustom::getImageUri
         // Support webp emojis by forcing every emoji to be webp
         patcher.instead<ModelEmojiCustom?>(
             "getImageUri",
             Long::class.javaPrimitiveType!!, Boolean::class.javaPrimitiveType!!, Int::class.javaPrimitiveType!!,
-        ) { param ->
-            "https://cdn.discordapp.com/emojis/${param.args[0]}.webp?size=${param.args[2]}&animated=${param.args[1]}"
+        ) { (_, id: Long, animated: Boolean, size: Int) ->
+            "https://cdn.discordapp.com/emojis/$id.webp?size=$size&animated=$animated}"
```
rushiiMachine: Extra brace at the end

rushiiMachine (changes requested):

PR #700, opened 2026-03-03
feat: injector rewrite & early init

rushiiMachine:
This completely replaces the old Aliucord initialization flow to allow for "early initialization" of the remotely downloaded Aliucord core bundle, meaning, that it is loaded as soon as `Application#onCreate` is invoked. This occurs when any background service/receiver/provider/activity starts, which now exposes a lot more potential modding surfaces such as fixing notifications, using custom gateway backends, or anything that requires replacing behavior that occurs early in the app's lifecycle.

The way this is implemented is as follows:
1. The Injector overrides a class that is initialized near the start of `com.discord.app.App#onCreate` (com.discord.utilities.lifecycle.ApplicationProvider) to invoke it's own internal initialization first
2. If manage/write external storage permissions are not granted, jump to the restoration flow
3. Check whether custom core is enabled and exists (this required permissions)
	- If granted, copy to internal code cache
4. If not using custom core, check that the official core is downloaded.
	- If not downloaded, jump to the restoration flow
5. Add either the offical or custom core to classloader
6. Find the appropriate entrypoint in loaded core
7. If a now-designated "legacy" (with `preInit`/`init` entrypoint methods that do not support early-init) was loaded,
	7.1.  Hook Discord's `AppActivity#onCreate`, initializing LSPlant in the process
	7.2. Once Activity was loaded, invoke `preInit`/`init` one after another (dumb)
8. If a new core was loaded, (for now, it's just a concept API) invoke the `onApplicationCreate` entrypoint.
9. The core handles initializing everything else, including hooking `AppActivity#onCreate`

Restoration flow:
1. Hook Discord's `AppActivity#onCreate`, initializing LSPlant in the process, and wait for the user to launch the activity
	- This is necessary because otherwise Aliucord cannot request permissions without an activity, and since the Injector may have been called due to a service starting, rather than the user launching Aliucord
2. If permissions are not granted, request permissions and wait without blocking main thread, letting stock Discord initialize
	- If not granted, warn user and let stock Discord continue functioning
10. Check whether custom core is enabled and exists (this requires the permissions that were granted)
	- If custom core is enabled, immediately restart the process
	- If not enabled, continue installing core
11. Fetch remote Aliucord core build data
12. If remote build requires a higher Kotlin stdlib than is available, ask user to update Aliucord with Manager and do not download core
	- Auto-redirecting to Manager is left for a future PR
13. Download Aliucord core
14. Restart the process


Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+                    return@Thread
+
+                if (!isUsingCustomCore() && !installCore())
+                    return@Thread
 
-        Logger.d("Starting Aliucord core...")
-        onApplicationInit.invoke(null, appCtx)
-    } catch (_: NoSuchMethodException) {
-        startLegacyAliucord()
+                restartAliucord()
```
cillynder, line 170: I don't think this would work outside the main thread(?) Since it starts a new activity

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+            throw IllegalStateException("Cannot hook AppActivity.onCreate when it was already invoked!")
+
+        try {
+            var unhook: XC_MethodHook.Unhook? = null
+            unhook = XposedBridge.hookMethod(
+                AppActivity::class.java.getDeclaredMethod("onCreate", Bundle::class.java),
+                object : XC_MethodHook() {
+                    override fun beforeHookedMethod(param: MethodHookParam) {
+                        unhook!!.unhook()
+                        activityInitialized.set(true)
```
cillynder, line 373: It's probably impossible for hookActivityOnCreate to run twice, but this *technically* isn't safe for race conditions.
```suggestion
                        if (activityInitialized.getAndSet(true)) return
```

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
-    conn.setRequestProperty("User-Agent", "Aliucord Injector/${BuildConfig.VERSION} (https://github.com/Aliucord/Aliucord)")
-    conn.useCaches = false
-
-    val data = JSONObject(conn.inputStream.bufferedReader().readText())
-    return BuildData(
-        discordVersion = data.getInt("versionCode"),
-        kotlinVersion = data.optString("kotlinVersion", "1.5.21"),
-    )
+        if (activityInitialized.get())
+            throw IllegalStateException("Cannot hook AppActivity.onCreate when it was already invoked!")
```
cillynder: Should add braces

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+    private fun disableCustomCore(): Boolean {
+        try {
+            if (!externalSettingsFile.exists()) return true
 
-                downloaded += bytes
-                bytes = stream.read(buffer)
-            }
+            val settings = JSONObject(externalSettingsFile.readText())
+            if (!settings.optBoolean(ALIUCORD_FROM_STORAGE_KEY, false))
+                return true
```
cillynder: Should add braces

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+            Logger.d("Failed to read external Aliucord settings, skipping custom core check...", e)
+            return false
         }
-    } catch (e: Throwable) {
-        Logger.errorToast(ctx, "Failed to download latest Aliucord core!", e)
-        false
+
+        val customCoreEnabled = settings.optBoolean(ALIUCORD_FROM_STORAGE_KEY, false)
+        if (customCoreEnabled)
+            Logger.d("Loading external custom Aliucord core!")
```
cillynder: This logging shouldn't be happening here, or at least it should be changed to something other than "loading".

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+            onApplicationInit.invoke(null, appCtx)
+        } catch (_: NoSuchMethodException) {
+            startLegacyAliucord()
         }
 
-        JSONObject(settingsFile.readText())
-    } catch (e: Exception) {
-        Logger.d("Failed to read external Aliucord settings, skipping custom core check...", e)
-        return false
+        Logger.d("Finished starting Aliucord")
```
cillynder: Should be removed since if legacy core is used it technically hasn't started.

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+    private fun installCore(): Boolean {
+        return try {
+            val data = fetchBuildData()
+            Logger.d("Retrieved remote build data: $data")
+
+            val kotlinSemver = data.kotlinVersion
+                .split('.')
+                .map { it.toInt() }
+
+            if (data.discordVersion > com.discord.BuildConfig.VERSION_CODE ||
```
cillynder, line 225: I don't think there's any point in checking the discord version anymore

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,377 @@
...
+            Logger.w("Failed to disable hidden api restrictions")
+
+        if (!XposedBridge.disableProfileSaver())
+            Logger.w("Failed to disable profile saver")
 
-    if (!XposedBridge.disableProfileSaver())
-        Logger.w("Failed to disable profile saver")
+        pruneArtProfile(appCtx)
+    } catch (e: Exception) {
+        Logger.e("Failed to setup environment", e)
```
cillynder, line 55: Would this ever throw?

cillynder (changes requested):

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: The intent behind this flag wasn't to prevent the callback from being invoked (potentially an infinite amount of times), it was to prevent hookActivityOnCreate from hooking onCreate again even after the activity has already been made.

rushiiMachine: I could document this better

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: Yes, if aliuhook fails. It would just crash the app if the exception isn't caught

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: I would rather keep it here for consistency since it's checked everywhere

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: It does actually work on other threads surprisingly

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
     Logger.d("Started Aliucord Injector!")
 
-    if (!XposedBridge.disableHiddenApiRestrictions())
-        Logger.w("Failed to disable hidden api restrictions")
+    try {
+        if (!XposedBridge.disableHiddenApiRestrictions())
+            Logger.w("Failed to disable hidden api restrictions")
+
+        if (!XposedBridge.disableProfileSaver())
+            Logger.w("Failed to disable profile saver")
```
zt64:
```kt
if (!XposedBridge.disableHiddenApiRestrictions()) {
    Logger.w("Failed to disable hidden api restrictions")
}

if (!XposedBridge.disableProfileSaver())
    Logger.w("Failed to disable profile saver")
}
```

Injector/src/main/java/com/aliucord/injector/Updater.kt
```diff
@@ -0,0 +1,78 @@
...
+    Logger.d("Downloaded Aliucord core after ${System.currentTimeMillis() - startTime}ms")
+}
+
+/**
+ * A parsed model of the data available at [DATA_URL].
+ */
+internal data class BuildData(
+    var discordVersion: Int,
+    var kotlinVersion: String,
+)
```
zt64: Perhaps add a field that uses the `KotlinVersion` class for simplifying usage in Injector.kt

Injector/src/main/java/com/aliucord/injector/Updater.kt
```diff
@@ -0,0 +1,78 @@
...
+                }
+
+                downloaded += bytes
+                bytes = stream.read(buffer)
+            }
+        }
+    }
+    tmpFile.renameTo(outputFile)
+    Logger.d("Downloaded Aliucord core after ${System.currentTimeMillis() - startTime}ms")
+}
```
zt64, line 54: I believe [measureTime](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/measure-time.html) would work for this.

Injector/src/main/java/com/aliucord/injector/Updater.kt
```diff
@@ -0,0 +1,78 @@
...
+            while (bytes >= 0) {
+                out.write(buffer, 0, bytes)
+
+                val newProgress = (downloaded + bytes) / length.toFloat()
+                if (newProgress >= oldProgress + 0.1f) {
+                    oldProgress = newProgress
+                    Logger.d(String.format(Locale.ROOT,
+                        "Downloaded %.2f%% after %sms",
+                        newProgress * 100,
+                        System.currentTimeMillis() - startTime))
```
zt64: Use the [format](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/format.html) extension function instead

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
 
 /**
  * External settings key controlling whether custom core is enabled.
  */
 private const val ALIUCORD_FROM_STORAGE_KEY = "AC_from_storage"
 
-private val initialized = AtomicBoolean(false)
-private val legacyInitialized = AtomicBoolean(false)
+private val applicationInitialized = AtomicBoolean(false)
+private val activityInitialized = AtomicBoolean(false)
```
zt64, line 33: An enum class might be more suited for this to explicitly track where in the initialization flow is rather than relying on two variables. AtomicReference can be used to hold an enum.

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
+     * A copy of [externalCustomCoreFile], to prevent corruption while Aliucord is already running.
+     * This is inaccessible to users and stored in internal cache.
+     */
+    private val internalCustomCoreFile = appCtx.codeCacheDir.resolve("Aliucord.custom.zip")
+
+    /**
+     * This is invoked when [App.onCreate] is called, triggering a possible early initialization of the Aliucord core
+     * if permissions have been granted and the core has already been downloaded during a prior launch.
+     */
+    @Throws(Exception::class)
```
zt64: Does `@Throws(Exception::class)` actually do anything? Exception is a base class for lots of other exceptions.

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
+        return try {
+            val data = fetchBuildData()
+            Logger.d("Retrieved remote build data: $data")
+
+            val kotlinSemver = data.kotlinVersion
+                .split('.')
+                .map { it.toInt() }
+
+            if (data.discordVersion > com.discord.BuildConfig.VERSION_CODE ||
+                !KotlinVersion.CURRENT.isAtLeast(kotlinSemver[0], kotlinSemver[1], kotlinSemver[2])
```
zt64: Adding KotlinVersion field to the BuildData, would simplify usage here so it could be like `KotlinVersion.CURRENT < buildData.kotlinVersion`

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
+        val useCustomCore = isUsingCustomCore()
 
-    try {
-        // Delete old custom cores copied to code cache if exist
-        if (!customCore) {
+        // Delete old custom cores copied to code cache if they exist
+        if (!useCustomCore) {
             internalCustomCoreFile.delete()
+        } else {
+            Logger.d("Using external custom Aliucord core!")
```
zt64: `external custom`, external core already implies custom so I would remove one of those adjectives

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
-                )
-            })
-
-            try {
-                if (!result.get())
-                    throw Exception("Aliucord core installer failed!")
-            } finally {
-                es.shutdown()
-            }
+            restoreCoreFlow()
```
zt64, line 119:
```kt
        } else if (!internalCoreFile.exists()) {
            // Download new stable core
            restoreCoreFlow()
```

Injector/src/main/java/com/aliucord/injector/Injector.kt
```diff
@@ -6,302 +6,379 @@
...
-                bytes = stream.read(buffer)
+            JSONObject(externalSettingsFile.readText())
+        } catch (e: Exception) {
+            // This may include permission errors
+            Logger.d("Failed to read external Aliucord settings, skipping custom core check...", e)
+            return false
+        }
+
+        return settings.optBoolean(ALIUCORD_FROM_STORAGE_KEY, false)
+    }
```
zt64, line 331: This would be better as a lazy variable with the same name

zt64 (changes requested):

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: I thought about this and I think it adds more overhead and complexity that just isn't worth it imo

Injector/src/main/java/com/aliucord/injector/Updater.kt
rushiiMachine: I would still have to keep the code almost the same, since that doesn't help while it's still running.
It would also increase the nesting by one

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: Can't do since this may be called multiple times throughout the initialization/restoration flow, at which points the result of this may change (ie, after granting permissions).

Injector/src/main/java/com/aliucord/injector/Injector.kt
rushiiMachine: That comment does reference the `else if` statement as a whole, also I think this visually looks worse 
image

PR #702, opened 2026-03-13
fix(Patches): patch BaseSticker too

Archimedes9500:
`StickerView` uses `BaseSticker` which is currently unpatched, so it (only sometimes) explodes when trying to render stickers

I'm gonna test this in a second so it's a draft


Archimedes9500:
Turns out this error is caused by magic and not this

PR #716, opened 2026-05-02
fix(PluginUpdater): Failing URL transformation

Ushie:
Fixes https://discord.com/channels/811255666990907402/811262084968742932/1500121054503243867

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/updater/PluginUpdaterSource.kt
+++ b/Aliucord/src/main/java/com/aliucord/updater/PluginUpdaterSource.kt
@@ -17,7 +17,7 @@ internal class PluginUpdaterSource {
 
     /**
      * A TTL cache of updater data from plugin repositories.
-     * This maps the **original** plugin update info url to the fetched json.
+     * This maps the **original** plugin update info url to the fetched JSON.
      */
     private val cachedRepoInfo = ConcurrentHashMap<String, RepoBuildInfo>()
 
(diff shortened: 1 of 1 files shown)


Ushie:
I noticed that even after the rewrite, the plugin updating system is still plugin-first instead of author/repo first, you could possibly gather the repos of the installed plugins ahead of time and then running the update checks for the installed plugins

Otherwise preperations like this 
```kt
        val realUpdateInfoUrl = updateInfoUrl.replace(
            "https://raw\\.githubusercontent\\.com/(.+?)/(.+?)/(.+)".toRegex(),
            "https://cdn.jsdelivr.net/gh/$1/$2@$3"
        )
```
run for every single plugin even if the exact same string has already been transformed 100 times, on my install this is probably being done over 200-300 times instead of just like 20 times

rushiiMachine:
> run for every single plugin even if the exact same string has already been transformed

I don't really see this as a problem, it's not performance sensitive (not that's it's even slow in the first place) and adding more state would complicate the updater even more imo

rushiiMachine:
Wait so what is even being fixed here, what was the bug?

Ushie:
The transformation doesn't make any changes, it was always returns back raw.githubusercontent.com

I didn't look into WHY the existing code didn't work, but my assumption is that replaceFirst doesn't take in regex it's for things like prefixes instead

"hi ushie hi".replaceFirst("hi", "hello") // hello ushie hi

is probably the usecase for this, because the replaceFirst takes in a string not a regex, replace does take regex instead

Ushie:
image

This is what was being used before, which doesn't do regex

rushiiMachine:
insane when i refactored the updater into kotlin apparently i didn't take this into account

PR #718, opened 2026-05-04
fix: support autocomplete entries with the same name

cillynder:
Most notable when two people have the same nickname/display name — only one would show up as the autocompletes are backed by a treeset and the comparator only compares the names. This PR makes the comparator also check for username[#discrim] to ensure all entries are shown

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -22,6 +22,7 @@ import com.aliucord.utils.ReflectUtils
 import com.aliucord.utils.ViewUtils.findViewById
 import com.aliucord.utils.accessField
 import com.aliucord.wrappers.ChannelWrapper.Companion.id
+import com.aliucord.wrappers.GuildRoleWrapper.Companion.name
 import com.aliucord.wrappers.embeds.MessageEmbedWrapper
 import com.discord.api.channel.Channel
 import com.discord.api.message.embed.EmbedField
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -355,6 +359,32 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+            // For roles
+            if (p1 is RoleAutocompletable && p2 is RoleAutocompletable) {
+                param.result = compareValuesBy(
+                    p1, p2,
+                    { it.role.name }, // Compare by name first
+                    { it.role.id }, // Then compare by id
+                )
+            }
+            // For users
+            if (p1 is UserAutocompletable && p2 is UserAutocompletable) {
```
rushiiMachine: else if or early return instead

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -355,6 +359,32 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+                param.result = compareValuesBy(
+                    p1, p2,
+                    { it.nickname ?: it.user.username }, // Compare by nickname/display name first
+                    { it.user.username + UserUtils.INSTANCE.getDiscriminatorWithPadding(it.user) }, // Then compare by username[#discrim] (always unique)
```
rushiiMachine: Instead of concatenating, why not add an extra comparison to compare by discrim last?

rushiiMachine (changes requested):

rushiiMachine (approved):

PR #721, opened 2026-05-06
fix: memory leak in fragment manager and rxjava

Canny1913:
Fixes some memory leak issues caused by faulty base app code.


Canny1913:
While this is ready for review, the leaks will not go away until users disable `DedicatedPluginSettings` plugin.
I can add a function that automatically disables it. (similar to the one in `PluginDownloader` plugin)

rushiiMachine (approved):

Issue #364, opened 2023-06-18
Servers with forum channels are always unread


bart7782:
### Discord Account

bprt

### What happens when the bug or crash occurs?

Any server with a forum channel will frequently be displayed as unread (white dot). This happens even though every (non muted) channel and forum post has been read. 

Even if i click "Mark as Read" in the server options then the server will still show as unread.

When i restart Aliucord the server seems to be read again. But if i switch to the server and back, then the server will show up as unread again.

### What is the expected behaviour?

The expected behaviour would be to show up as Read. (Without a white dot next to the server)

### How do you recreate this bug or crash?

1. Go to a server that has a forum channel
2. Make sure the forum channel is not muted
3. Click "Mark as Read" in the server options and read every forum post.
4. Switch so a different server and the server with a forum channel will still show as unread

### Request Agreement

- [X] I did indeed check to make sure the bug or crash report is applicable!

BeeFox-sys:
the issue is twofold:
forum channels do not show if there are unread messages in them, but do cause the server to show there are unread messages
marking a server as read does not effect forum posts

aubymori:
bump, this is really annoying for me

QuinceTart10:
Same issue here, it's really annoying.

aubymori:
bump please for the love of god fix this it's the most annoying thing ever

Icyy1:
bump please fix

aubymori:
UPDATE: I figure out what causes this. Servers with unopened forum posts do this. Any without don't.

rushiiMachine:
Fixed in #722

aubymori:
oh my god finally thank you

Canny1913:
> oh my god finally thank you

have funnn (took me 1 hour to figure it out)

aubymori:
> > oh my god finally thank you
> 
> have funnn (took me 1 hour to figure it out)

Unrelated, but is there any update on animated WEBP fix? The PR for it (which you made) is closed at the moment...

Canny1913:
> > > oh my god finally thank you
> > 
> > 
> > have funnn (took me 1 hour to figure it out)
> 
> Unrelated, but is there any update on animated WEBP fix? The PR for it (which you made) is closed at the moment...

i closed it since im focusing on other things right now but i will finish it eventually

Canny1913:
> > > oh my god finally thank you
> > 
> > 
> > have funnn (took me 1 hour to figure it out)
> 
[...]

Issue #379, opened 2023-08-31
Can't clear notifications on Forums easily


TonyGamer:
### Discord Account

Tonì#7947

### What happens when the bug or crash occurs?

When trying to clear notifications on forums, the notifications *appear* to be cleared, but the server still displays unread notifications, and if you enter into the forums, the posts still display the "new" test, as if they have unread messages.

### What is the expected behaviour?

I would expect (and this is the behavior in the official app) that marking a server or forum as read would mark all messages within the forums as read, thus removing the "unread messages" icon from the server. Doing the same in the official app causes the forums to be marked read correctly, both in Discord *and* Aliucord

### How do you recreate this bug or crash?

1. Mark server with unread forum messages as read using the tap and hold context menu
2. Click off the server to somewhere else
3. Notice that the server is still marked as containing unread messages

### Request Agreement

- [X] I did indeed check to make sure the bug or crash report is applicable!

TonyGamer:
For more information, I am running 126.21 - Stable, Aliucord 2CF5CE8

FlitPix:
Related to #364

nosut:
Having the same issue. Would rather avoid muting the channel completely as well.

rushiiMachine:
Fixed in #722

PR #722, opened 2026-05-06
fix: unread indicator in guilds with forum channels

Canny1913:
Fixes unread indicator false trigger in guilds with forum channels.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -29,6 +29,7 @@ import com.discord.api.permission.Permission
 import com.discord.models.domain.emoji.ModelEmojiCustom
 import com.discord.models.domain.emoji.ModelEmojiUnicode
 import com.discord.rtcconnection.socket.io.Payloads.Protocol.ProtocolInfo
+import com.discord.stores.StoreReadStates
 import com.discord.stores.StoreSlowMode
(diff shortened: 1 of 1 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -421,7 +423,27 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+            Map::class.java,
+            Map::class.java,
+            Map::class.java,
+            Long::class.javaPrimitiveType!!,
+            Map::class.java,
+            Map::class.java,
+        )
+        { param ->
+            // Only check joined forum posts instead of all public posts
+            val forumThreadsMap = param.args[7] as Map<Long, Channel>
```
rushiiMachine, line 440: If this is a mutable map, could you mutate it directly instead of recreating a filtered one perhaps?

rushiiMachine (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
Canny1913: The retrieved map is 'forumThreadsByIdSnapshot' from StoreThreadsActive, which is mutable but would certainly break other things if mutated.

Canny1913:
Forgot to add
closes #364 and #379

PR #724, opened 2026-05-06
fix: Only check version equality for plugin updates

### LINKED ISSUE #723: PluginUpdater breaks upon encountering plugins with invalid semver versions
### Tell us all about it.

Unfortunately people can't stick to normal versions so here we are.

- Plugin updater checks for plugin updates
- Local & remote plugin version is invalid 
- Plugin updater says there are updates in notification but updater page displays nothing (likely filtered out)

Image
Image

### Request Agreement

- [x] I did indeed check to make sure the issue is original!

Ushie:
Closes #723

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.kt
+++ b/Aliucord/src/main/java/com/aliucord/updater/PluginUpdater.kt
@@ -88,7 +88,7 @@ internal object PluginUpdater {
                 }
 
                 // Plugin is already up-to-date
-                if (SemVer.parse(plugin.manifest.version) >= info.version)
+                if (plugin.manifest.version == info.version)
                     continue
 
                 updates += PluginUpdate(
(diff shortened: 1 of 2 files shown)


Ushie:
I checked the issue with notifications showing anyways but there's no reason for them to be shown

Ushie:
It goes through the exact lines of code that were failing to parse the semver and returning early instead, likely unrelated to this exact issue

rushiiMachine:
> I checked the issue with notifications showing anyways but there's no reason for them to be shown

Yeah I have no idea either, perhaps Juby's internet was just super slow atm and the page turned up blank

rushiiMachine:
I think we can go ahead with merging this

Ushie:
Not yet, a usecase where this would be annoying was mentioned, I'll convert it back to checking for greater than soon

Ushie:
Closing because we no longer plan to support improperly formatted versions, going forward only x.x.x versions will be supported

PR #728, opened 2026-05-11
fix(PluginUpdater): treat invalid local plugin versions as outdated

rushiiMachine:
This fixes updater issue on 6 obscure plugins that have had non-semver version strings.
All currently published plugins now follow semver version format


Ushie (approved):
Maybe installedVersion is a better name, but LGTM

PR #732, opened 2026-05-16
feat(PluginUpdater): Increase details in notifications

Archimedes9500:
Make plugin update notifications show updated plugin names again, as well as versions (so you know if there really is a new update or is it still the same broken Frecents update you've been ignoring the past week)

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/Main.java
+++ b/Aliucord/src/main/java/com/aliucord/Main.java
@@ -268,48 +268,83 @@ public final class Main {
         if (updates.isEmpty()) return;
 
         if (!PluginUpdater.isAutoUpdateEnabled()) {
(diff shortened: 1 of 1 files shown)


Ushie:
> is it still the same broken Frecents update you've been ignoring the past week

Maybe an option to ignore a specific plugin update would be good, what do you think?

Archimedes9500:
> Maybe an option to ignore a specific plugin update would be good, what do you think?

Yeah, I was working on that before the rewrite, but that probably should be a separate PR.
It should be easy to add that with how the logic is now.

Archimedes9500:
This is still relevant btw, I also tested it and it works, and should be ready for review.
Except you'll have to pull `patch-2` manually (or remake the PR) because github is fucking blind and won't let me reopen it

PR #733, opened 2026-05-16
feat(PluginUpdater): Increase details in notifications

Ushie:


Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -268,48 +269,80 @@ private static void checkForPluginUpdates() {
...
+                msg
+                    .append("**")
+                    .append(update.getPluginName())
+                    .append("**\u00A0(")//NBSP
+                    .append(update.getInfo().getVersion().toString())
+                    .append(")");
+            }
+
+            msg
+                .insert(0, String.format("Updates for %d plugins: ", updates.size()));
```
rushiiMachine: No reason to insert this afterwards

Also something like "Found %d plugin updates: update1, update2, update3, update4, and %d others" would be a better message. You shouldn't display **everything** in a single notification, that's not what it's for; when central migration occurs there might be over a hundred updates all at the same time.

Aliucord/src/main/java/com/aliucord/Main.java
rushiiMachine, line 1: Some changes are improperly formatted, reformat in android studio

Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -268,48 +269,80 @@ private static void checkForPluginUpdates() {
...
+                    .append(")");
             } else {
+                if(failed > 0) failedMsg.append(", ");
                 failed++;
+                failedMsg
+                    .append("**")
+                    .append(update.getPluginName())
+                    .append("**\u00A0(")//NBSP
+                    .append(update.getInfo().getVersion().toString())
+                    .append(")");
```
rushiiMachine: This is quite ugly and repetitive, also constructed way too early. Same snippet of code 3 times.

Aliucord/src/main/java/com/aliucord/Main.java
```diff
@@ -268,48 +269,80 @@ private static void checkForPluginUpdates() {
...
             }
         }
 
         var notification = new NotificationData()
             .setTitle("Updater");
-        if (failed > 0) {
+        if (failed == 0) {
+            succeededMsg
+                .insert(0, String.format("Automatically updated %s plugins: ", succeeded));
```
rushiiMachine: No reason to insert at the beginning, just construct the entire message here...

rushiiMachine (changes requested):

rushiiMachine:
I think if you rewrote this instead of working off of archi's code it would be easier

Aliucord/src/main/java/com/aliucord/Main.java
Archimedes9500: > No reason to insert this afterwards

The reason is so I can easily check if I'm adding first plugin or not without storing extra state
Does it really matter if it's inserted before or after though

> Also something like "Found %d plugin updates: update1, update2, update3, update4, and %d others" would be a better message.

I don't see a point in repeating the count twice, and the reason why I placed it at the start and not the end is because the end will be ellipsed, so displaying it there would require measuring text length to ensure it doesn't get cut off

> You shouldn't display **everything** in a single notification

It cuts off after 3 lines, so it doesn't actually display everything

Aliucord/src/main/java/com/aliucord/Main.java
Archimedes9500: > This is quite ugly and repetitive, also constructed way too early. Same snippet of code 3 times.

Idk about ugly, I guess it could be a single `String::format`.

I don't really understand what do you mean by "too early" though, I do it at the point of iteration, same as before (except unlike before, I don't just collect names and then `joinToString` anyway lmao)

The repetition is also the same as before (all 3 messages are constructed separately), except now they are consistently constructed the same way instead of having random differences for no reason whatsoever

Aliucord/src/main/java/com/aliucord/Main.java
Archimedes9500: ??? how
Or why rather
It already has to loop over all updates, why would I loop here **again** just to concat it?

rushiiMachine: Modifying state throughout multiple code flow paths makes it much more difficult to reason about what is actually happening

Archimedes9500: > Modifying state throughout multiple code flow paths makes it much more difficult to reason about what is actually happening

Fair enough, I did it to keep it consistent with regular `msg` which doesn't track state, but I guess it doesn't have to be the same

But also this is kind of how it was before too, just more cursed:
https://github.com/Aliucord/Aliucord/blob/28f8dfdcc322a7de35598205be2a6ff1673c56c3/Aliucord/src/main/java/com/aliucord/Main.java#L284-L312

Archimedes9500: Nvm that's not possible
What about just:
```suggestion
                .setBody(MDUtils.render(String.format("Automatically updated %s plugins: ", succeeded)
                    + succeededMsg))
```

Aliucord/src/main/java/com/aliucord/Main.java
Archimedes9500: now better?

zt64:
> I think if you rewrote this instead of working off of archi's code it would be easier

Agreed

PR #709, opened 2026-04-18
fix: animated webp emojis and avatars

Canny1913:
This PR fixes animated emojis and avatars in WebP format by modifying hardcoded GIF checks in Discord and adding `animated` query support.

PR #736, opened 2026-05-19
fix: support animated webp resources

cillynder:
Fix animated webp rendering in various places by using correct file extensions, and inclusions of `animated=true` query parameter.

Supersedes #709

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -170,15 +170,14 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
 
         // Fixup methods that add a ?size= to the url, to replace them with &size instead since
         // we already have ?animated=... above
-        @Suppress("LocalVariableName")
         run {
-            val String = String::class.java
-            val Long = Long::class.javaObjectType
-            val Integer = Int::class.javaObjectType
-            val boolean = Boolean::class.javaPrimitiveType!!
-            patchIconU("getForGuild", Long, String, String, boolean, Integer)
-            patchIconU("getBannerForGuild", Long, String, Integer, boolean)
-            patchIconU("withSize", String, Integer)
+            patchIconU("getForGuild",
+                Long::class.javaObjectType, String::class.java, String::class.java,
+                Boolean::class.javaPrimitiveType!!, Int::class.javaObjectType)
+            patchIconU("getBannerForGuild",
+                Long::class.javaObjectType, String::class.java, Int::class.javaObjectType,
+                Boolean::class.javaPrimitiveType!!)
+            patchIconU("withSize", String::class.java, Int::class.javaObjectType)
         }
         patcher.after<GuildListViewHolder.GuildViewHolder>(
             "configureGuildIconImage",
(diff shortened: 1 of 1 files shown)


cillynder:
Found a bug where clicking on a server icon for the first time briefly causes it to flicker, and animated server icons no longer work. Looking into it.

rushiiMachine (approved):

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -141,36 +145,107 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+                ?.takeIf { (it.path?.endsWith(".gif") == true && animated) || it.path?.endsWith(".webp") == true }
                 ?: return@after
 
-            val filteredQueryKeys = uri.queryParameterNames.filter { it != "format" }
+            val filteredQueryKeys = uri.queryParameterNames.filter { it != "format" && it != "animated" }
 
             val newUri = uri.buildUpon()
                 .clearQuery()
                 .apply { filteredQueryKeys.forEach { appendQueryParameter(it, uri.getQueryParameter(it)) } }
+                .apply { appendQueryParameter("animated", animated.toString()) }
```
zt64: `appendQueryParameter` returns the builder instance, so there's no need to use another `apply`

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -141,36 +145,107 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+        // Animate webp properly in fullscreen media view
+        patcher.before<WidgetMedia>(
+            "getFormattedUrl",
+            Context::class.java, Uri::class.java,
+        ) { (param, _: Context, uri: Uri) ->
+            if (uri.path?.endsWith(".webp") == true)
+                param.result = uri
+                    .buildUpon()
+                    .appendQueryParameter("animated", "true")
+                    .toString()
```
zt64: I would add some brackets to this if statement, looks a little weird to read at first

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -141,36 +145,107 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+        }
+
+        // Fixup methods that add a ?size= to the url, to replace them with &size instead since
+        // we already have ?animated=... above
+        @Suppress("LocalVariableName")
+        run {
+            val String = String::class.java
+            val Long = Long::class.javaObjectType
+            val Integer = Int::class.javaObjectType
+            val boolean = Boolean::class.javaPrimitiveType!!
```
zt64: For conciseness it would be better to name these `<class>Class` like `stringClass`

zt64 (changes requested):
Looks mostly good just a few changes needed

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
rushiiMachine: alternatively, you could just inline each one, it doesn't save that much space
i think we're all used to the ::class.java and ::class.java.primitiveJavaTime!! by now

zt64 (approved):

rushiiMachine:
@cillynder mergee

aubymori:
^

PR #740, opened 2026-05-28
feat: Improve plugin updater

### LINKED ISSUE #739: [Updater] Aliucord can fetch updates more than once on Updater screen

omegasunkey

### What’s the issue?

You can click the fetch updates button inside Updater screen while Updater is still fetching updates, which cause update entry duplication in the screen.

As there is no indicator of updater fetching updates, user would be confused and manually trigger update fetching, then get more confused at this.

It should not re-run fetching process

### How can we reproduce it?

1. Enter updater screen by either clicking notification or by clicking the Updater entry in settings
2. Click the fetch updates button while the screen is empty
3. Entry duplication happens when they both finalize fetching

https://github.com/user-attachments/assets/1e7743fd-b92f-40b1-8a8e-7190acf73efd

### Android Version

Android 13

### Device & OS Info

Samsung Galaxy A30, Arrow OS for Android 13

### Crash Log (if applicable)

_No response_

### Before Submitting

Ushie:


Ushie:
I managed two routes for the parcel route for notifs, I could make plugin a lateinit inside the class and set it right after, which worked, but opening the notification immediately after opening the app seems to be faster than that because it doesnt seem to be populated at that point

the other route is just making it a getter to get it from PluginManager but that means everytime its used its a call to plugin manager

Ushie:
Eitherway it feels like I'm compromising for the sake of making it a parcel just for startup notifications, is there really no way to pass data directly to a screen without intents?

Ushie:
Ready for review

Ushie:
🥹

PR #746, opened 2026-06-08
fix: QR & Token login

### LINKED ISSUE #730: QR code login no longer works

cutie_royce

### What’s the issue?

if 2FA is enabled on account and you try to login with QR from computer, it will vaguely say "Can't find this computer". ~~In the modern app, it seems logging in with QR now requires mandatory 2FA verification.~~ (It seems this is now affecting every user even if they have 2FA off?)

Can login without issue

### How can we reproduce it?

1. Make sure 2FA is enabled on the account
2. Try to sign in on the target computer with QR code
3. Try to accept it, it will say "can't find this computer"

### Android Version

Android 12 SDK 32

### Device & OS Info

N/A

### Crash Log (if applicable)

_No response_

### Before Submitting

secp192k1:
horrible..
~~also it LAGS when WAITING because NO loading~~
update: NO lag and HAS loading indications

works and logins 👍👍

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -7,11 +7,14 @@ import android.content.Intent
 import android.content.pm.PackageManager
 import android.net.Uri
 import android.os.*
+import android.view.MotionEvent
 import android.view.View
 import android.view.WindowInsetsAnimation
 import android.widget.TextView
 import androidx.core.content.ContextCompat
+import androidx.fragment.app.FragmentActivity
 import androidx.fragment.app.FragmentManager
+import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import androidx.viewbinding.ViewBinding
@@ -54,6 +57,7 @@ import com.discord.utilities.rest.RestAPI
 import com.discord.utilities.time.ClockFactory
 import com.discord.utilities.time.NtpClock
 import com.discord.utilities.viewbinding.FragmentViewBindingDelegate
+import com.discord.utilities.view.extensions.RecyclerViewExtensionsKt
 import com.discord.widgets.channels.list.*
 import com.discord.widgets.chat.input.*
 import com.discord.widgets.chat.input.autocomplete.adapter.ChatInputAutocompleteAdapter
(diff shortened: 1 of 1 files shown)


secp192k1:
Also it fixes #730

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
```diff
@@ -24,20 +30,202 @@
...
+    }
+
+    private static void doFinishRemoteAuth(AppFragment host, String handshakeToken, String mfaJwt) {
+        try {
+            Log.i(TAG, "Finish RA start. Handshake Token: " + handshakeToken);
+            String body = new JSONObject()
+                .put("handshake_token", handshakeToken)
+                .toString();
+
+            String response = discordPost("/users/@me/remote-auth/finish", body, mfaJwt);
```
dolfies: i feel like it would be significantly better to add proper mfa support to the existing logic instead of reimplementing everything

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
```diff
@@ -24,20 +30,202 @@
...
+            if (response.trim().isEmpty()) {
+                Log.i(TAG, "QR login success: 204 empty");
+                Utils.showToast("Logged in!", false);
+                return;
+            }
+
+            JSONObject json = new JSONObject(response);
+            int code = json.optInt("code", -1);
+            if (code == 60003) {
+                String ticket = json.getJSONObject("mfa").getString("ticket");
```
dolfies: totp is not the only mfa method (at least sms/backup codes are already supported by kotler)
for users without mfa, this will probably prompt for password instead

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
secp192k1: added & working on it

rushiiMachine:
Is there a reason this is written in Java? It's quite difficult to read

secp192k1:
> Is there a reason this is written in Java? It's quite difficult to read

Couple reasons, primary one being since `QrLogin` depends on `TokenLogin` and its written in Java, it would make sense, the core plugins are a mix of Kotlin and Java so I wouldn't think it would be the end of the world

zt64:
> > Is there a reason this is written in Java? It's quite difficult to read
> 
> Couple reasons, primary one being since `QrLogin` depends on `TokenLogin` and its written in Java, it would make sense, the core plugins are a mix of Kotlin and Java so I wouldn't think it would be the end of the world

We wanna migrate everything to Kotlin, not further make it even more of a mixed mess

secp192k1:
> > > Is there a reason this is written in Java? It's quite difficult to read
> > 
> > 
> > Couple reasons, primary one being since `QrLogin` depends on `TokenLogin` and its written in Java, it would make sense, the core plugins are a mix of Kotlin and Java so I wouldn't think it would be the end of the world
> 
> We wanna migrate everything to Kotlin, not further make it even more of a mixed mess

chop chop @zt64

rushiiMachine:
I'm confused, why did you interleave QR login fix that isn't a coreplugin with a rewritten TokenLogin plugin that has nothing to do with QR code login?

secp192k1:
@rushiiMachine because zeet is evil and made me rewrite it in kotlin and since there is a bit of QR logic in token login (like patches) might as well rewrite both while i was at it

secp192k1:
closed because i renamed the damn branch bruh

PR #751, opened 2026-06-13
feat(TokenLogin): UI feedback

### LINKED ISSUE #725: TokenLogin plugin should validate token
### Tell us all about it.

Accidentally copied token together with a newline, and didn't realize since the hidden chars made the input show as one line.
No error shown anywhere, it just accepted the input and then Aliucord just failed to load anything until I realized in logs that the Authorization header was invalid.

It should auto trim whitespace on both ends and also perform token format validation (and maybe even testing the token itself too) prior to accepting and storing it.

### Request Agreement

- [x] I did indeed check to make sure the issue is original!

secp192k1:
## Whats this

Logging in with a token previously gave absolutely 0 user feedback, so when trying to login with an invalid/expired token, it just silently fails with no visible change, leaving you clueless

Now:
- The token is validated by making a request to `/users/@me` before running the login dispatch
- Results surface via toasts: empty input, invalid token (status code and response included), and other errors
- While the token is being validated, the login button turns into a loading spinner and its state becomes disabled (before you could spam click while the validation was running)

## Preview

https://github.com/user-attachments/assets/5b951871-54a2-440b-a834-87beda83efb4

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
@@ -11,6 +11,7 @@ import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 
+import com.aliucord.Http;
 import com.aliucord.Utils;
 import com.aliucord.entities.CorePlugin;
 import com.aliucord.patcher.Patcher;
@@ -23,11 +24,13 @@ import com.discord.models.domain.auth.ModelLoginResult;
 import com.discord.stores.StoreAuthentication;
 import com.discord.stores.StoreStream;
 import com.discord.utilities.view.extensions.ViewExtensions;
+import com.discord.views.LoadingButton;
 import com.discord.widgets.auth.WidgetAuthLanding;
 import com.google.android.material.button.MaterialButton;
 import com.google.android.material.textfield.TextInputLayout;
 import com.lytefast.flexinput.R;
 
+import java.io.IOException;
 import java.util.*;
 
 import kotlin.Unit;
@@ -39,6 +42,8 @@ public final class TokenLogin extends CorePlugin {
     }
 
     public static class Page extends AppFragment {
+        private LoadingButton loginButton;
+
         public Page() {
             super(Utils.getResId("widget_auth_login", "layout"));
         }
@@ -48,6 +53,8 @@ public final class TokenLogin extends CorePlugin {
             super.onViewBound(view);
 
             LinearLayout v = view.findViewById(Utils.getResId("auth_login_container", "id"));
+            if (v == null) return; // prevent crash if layout change
+
             v.removeViewAt(1); // remove email input
             v.removeViewAt(2); // remove forgot password
             v.removeViewAt(2); // remove use a password manager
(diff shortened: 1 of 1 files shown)


Ushie:
lol nice group name

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
```diff
@@ -56,23 +63,60 @@ public void onViewBound(View view) {
...
+                        StoreStream.getAuthentication(),
+                        new ModelLoginResult(trimmedToken.startsWith("mfa."), null, trimmedToken, null, new ArrayList<>())
+                    );
+                } catch (Http.HttpException e) {
+                    setLoading(false);
+                    Utils.showToast("Invalid token: " + e.statusCode + ": " + e.statusMessage);
+                } catch (IOException e) {
+                    setLoading(false);
+                    Utils.showToast("Failed to verify token: " + e.getMessage());
+                }
```
zt64, line 107: You can use a `finally` block to set the loading state

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
```diff
@@ -56,23 +63,60 @@ public void onViewBound(View view) {
...
+
+        private void setLoading(boolean state) {
+            if (loginButton == null) return;
+            if (state) {
+                loginButton.setIsLoading(true);
+                loginButton.setEnabled(false);
+            } else Utils.mainThread.post(() -> {
+                loginButton.setIsLoading(false);
+                loginButton.setEnabled(true);
+            });
```
zt64:
```java
            if (state) {
                loginButton.setIsLoading(true);
                loginButton.setEnabled(false);
            } else {
                Utils.mainThread.post(() -> {
                    loginButton.setIsLoading(false);
                    loginButton.setEnabled(true);
                });
            }
```

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
```diff
@@ -56,23 +63,60 @@ public void onViewBound(View view) {
...
+            String trimmedToken = token.toString().trim();
+            if (trimmedToken.isEmpty()) {
+                Utils.showToast("Token cannot be empty");
+                return;
+            }
+            setLoading(true);
+            Utils.threadPool.execute(() -> {
+                try (Http.Request req = Http.Request.newDiscordRNRequest("/users/@me")
+                        .setHeader("Authorization", trimmedToken)
+                        .setRequestTimeout(10000)) {
```
zt64, line 95: Probably in a separate PR, timeout should be made a constant variable in the Http class rather than using a magic number

zt64 (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java
secp192k1: Should that be const val or use the existing:
```java
public Request setRequestTimeout(int timeout) {
            conn.setConnectTimeout(timeout);
            conn.setReadTimeout(timeout);
            return this;
        }
```

secp192k1: Also this number is across various places

image

rushiiMachine:
Have you verified that this also trims newlines automatically (ones that are pasted in)?

secp192k1:
> Have you verified that this also trims newlines automatically (ones that are pasted in)?

Yes

https://github.com/secp192k1/Aliucord/blob/85ece3dfd7340e376a578c2560afbf1cb32ff8f9/Aliucord/src/main/java/com/aliucord/coreplugins/TokenLogin.java#L86

rushiiMachine:
fwiw the old implementation "trimmed" too but didn't get rid of newlines

rushiiMachine (approved):

secp192k1:
> fwiw the old implementation "trimmed" too but didn't get rid of newlines

I know exactly what you mean, its when pasting text it adds a newline, even though you cant add newlines on the keyboard
Even with those pasted newlines, it still removes the, so this is only a UI/UX thing

PR #755, opened 2026-06-18
feat: friend request message

Canny1913:
Adds support for Friend Request Accepted message type.
Plugin can be extended to include other missing message types.

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/NewMessages.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/NewMessages.kt
@@ -1,0 +1,93 @@
+/*
+ * This file is part of Aliucord, an Android Discord client mod.
+ * Copyright (c) 2025 Juby210 & Vendicated
+ * Licensed under the Open Software License version 3.0
+ */
+
+package com.aliucord.coreplugins
+
+import android.content.Context
+import android.text.SpannableString
+import android.text.SpannableStringBuilder
+import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.RecyclerView
+import com.aliucord.Main
+import com.aliucord.entities.CorePlugin
+import com.aliucord.patcher.*
+import com.aliucord.utils.DimenUtils
+import com.discord.stores.StoreStream
+import com.discord.utilities.spans.ClickableSpan
+import com.discord.widgets.chat.list.adapter.*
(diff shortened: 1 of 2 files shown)


cillynder (commented):
Does this handle the scenario where you're the one accepting the other's request? It would be saying "<you> accepted your friend request"
I'm not sure how that's supposed to look because I can't seem to get this message to generate, but otherwise code lgtm

cillynder:
Also while it's fine for now, this should be abstracted at one point to make it easy to add more similar message types

Polls result message should be moved here too when that happens

Aliucord/src/main/java/com/aliucord/coreplugins/NewMessages.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.discord.stores.StoreStream
+import com.discord.utilities.spans.ClickableSpan
+import com.discord.widgets.chat.list.adapter.*
+import com.discord.widgets.chat.list.entries.ChatListEntry
+import com.discord.widgets.chat.list.entries.MessageEntry
+import com.lytefast.flexinput.R
+
+internal class NewMessages : CorePlugin(Manifest("NewMessages")) {
+    init {
+        manifest.description = "Adds missing message types"
```
rushiiMachine, line 30: I think this should be marked required+hidden

rushiiMachine (changes requested):

Aliucord/src/main/java/com/aliucord/coreplugins/NewMessages.kt
Canny1913: Why hide?

rushiiMachine: Why would it not be? As I understand it this is a fix, so perhaps this should even be included in CoreFixes?

Canny1913: > Why would it not be? As I understand it this is a fix...

This is a backport since i ported a feature that did not exist in this client, so you still lack a reason to hide the plugin. But i agree with making it force-enabled.

> ...so perhaps this should even be included in CoreFixes?

Cilly's suggestion makes more sense since there are many more message types that could be added in the future.

rushiiMachine (approved):

PR #756, opened 2026-06-20
feat(PluginsList): Show authors for core plugins, rewrite visibility toggles

secp192k1:


Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -278,6 +296,14 @@ public void onUninstallClick(int position) {
         public static boolean filterCorePlugins(Plugin p) {
             return !(p instanceof CorePlugin);
         }
+
+        private static void setVisible(View v, boolean visible) {
+            v.setVisibility(visible ? View.VISIBLE : View.GONE);
+        }
+
+        private static boolean notBlank(String s) {
```
zt64: `isNotBlank` would be a better name for this function.

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -121,24 +121,42 @@ public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
...
+
+            setVisible(holder.card.settingsButton, hasSettings);
             holder.card.settingsButton.setEnabled(enabled);
-            holder.card.uninstallButton.setVisibility(p.__filename != null ? View.VISIBLE : View.GONE);
-            holder.card.repoButton.setVisibility(p.getManifest().updateUrl != null ? View.VISIBLE : View.GONE);
-            holder.card.changeLogButton.setVisibility(p.getManifest().changelog != null ? View.VISIBLE : View.GONE);
+            setVisible(holder.card.uninstallButton, hasUninstall);
+            setVisible(holder.card.repoButton, hasRepo);
+            setVisible(holder.card.changeLogButton, hasChangelog);
+            setVisible(holder.card.buttonLayout, hasSettings || hasUninstall || hasRepo || hasChangelog || hasDescription);
```
cillynder: I don't think this is necessary; it's already pretty much hidden if there's no buttons

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -121,24 +121,42 @@ public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
         public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
             Plugin p = data.get(position);
             Plugin.Manifest manifest = p.getManifest();
+            String name = p.getName();
+            boolean core = p instanceof CorePlugin;
```
cillynder: Probably just inline these

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -121,24 +121,42 @@ public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
...
+            holder.card.switchHeader.setButtonVisibility(toggleable);
+            // TODO: Add a toast "Cannot stop required coreplugin ..."
+            holder.card.switchHeader.l.b().setClickable(toggleable);
+            holder.card.descriptionView.setText(MDUtils.render(manifest.description));
+            setVisible(holder.card.descriptionView, hasDescription);
+
+            boolean hasSettings = p.settingsTab != null;
+            boolean hasUninstall = notBlank(p.__filename);
+            boolean hasRepo = notBlank(manifest.updateUrl);
+            boolean hasChangelog = notBlank(manifest.changelog);
```
cillynder: Because L148 shouldn't be necessary, just inline these into each usage

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -121,24 +121,42 @@ public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
         public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
             Plugin p = data.get(position);
             Plugin.Manifest manifest = p.getManifest();
+            String name = p.getName();
+            boolean core = p instanceof CorePlugin;
+
+            boolean enabled = PluginManager.isPluginEnabled(name);
+            boolean toggleable = !core || !((CorePlugin) p).isRequired();
+            boolean hasDescription = notBlank(manifest.description);
```
cillynder: Inline this too

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
cillynder: Please just use a `SpannableStringBuilder` for this

cillynder (changes requested):

zt64 (changes requested):
Also several variables could be inlined

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
secp192k1: It is, `setVisible` sets the child buttons to `GONE`, but `buttonLayout` has a 2dp padding so there will always be that padding at the very least, without that code, buttonless cards leave ~8dp empty strip at the bottom

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
secp192k1: Like this?
```java
String name = p.getName(); boolean core = p instanceof CorePlugin;
```

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
secp192k1: the code repetition kills me.. but yes

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
secp192k1: im actually fucking stupid

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
secp192k1: Here is a visual example
With the last line `setVisible(holder.card.buttonLayout, [...] `  (now `setVisible(holder.card.buttonLayout, [...]` commented out, it will have that extra 2dp padding at all times

image

Aliucord/src/main/java/com/aliucord/settings/Plugins.java
```diff
@@ -207,8 +223,8 @@ public Filter getFilter() {
         }
 
         private String getGithubUrl(Plugin plugin) {
-            return plugin
-                .getManifest().updateUrl.replaceFirst(
+            return Objects.requireNonNull(plugin
+                .getManifest().updateUrl).replaceFirst(
```
rushiiMachine: `updateUrl` is not guaranteed to exist. This would cause a crash (as I assume it already did) if no `updateUrl` exists but a changelog does (see other usage of this).

rushiiMachine (changes requested):

rushiiMachine (approved):

PR #757, opened 2026-06-20
perf: Core fixes

secp192k1:
## the fixing

**1. Profile banner color thread crash**
`CalledFromWrongThreadException` when opening a profile. Fresco runs on a background thread and calls `updateBannerColor` and `setBackgroundColor`, which area on the main thread

**2. DefaultStickers send NPE**
`NullPointerException` on `Collection.isEmpty()` while sending a message

**3. BottomSheet callback warning**
`BottomSheetBehavior.setBottomSheetCallback()`, used by `AppBottomSheet.onViewCreated`, is deprecated

**4. Storage perms dialog span crash**
`Utils.launchFileExplorer` builds a `SpannableStringBuilder` and calls `setSpan` at `text.indexOf(substring)`
When the label is missing, it throws an `IndexOutOfBoundsException` because `indexOf` == `-1`

**5. Limit message length to prevent freezes**
`EditText` has no character limit, so pasting a very large wall of text into it freezes the app completely

**6. Fix crash on "NEW! Text chat in voice" popup crash**
When you click on "Try it" right as the popup shows, the activity is `null`, you must wait a while for it to load in the background, then it becomes `com.discord.app.AppActivity$Call@...`

<details>
<summary>image preview of that popup</summary>

text chat in voice crash
</details>

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -197,6 +197,8 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
     private val GuildListViewHolder.GuildViewHolder.bindingGuild
         by accessField<WidgetGuildsListItemGuildBinding>()
 
+    private val guildIconUrls = WeakHashMap<View, String>()
+
     private fun patchIconU(name: String, vararg paramTypes: Class<*>) {
         patcher.patch(IconUtils::class.java.getDeclaredMethod(name, *paramTypes)) {
             it.result = (it.result as? String)
@@ -232,11 +234,22 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
             "configureGuildIconImage",
             Guild::class.java, Boolean::class.javaPrimitiveType!!,
         ) { (_, guild: Guild, animated: Boolean) ->
-            if (!guild.hasIcon()) return@after
+            val icon = bindingGuild.d
 
-            val size = IconUtils.getMediaProxySize(bindingGuild.d.layoutParams.height)
+            if (!guild.hasIcon()) {
+                guildIconUrls -= icon
+                return@after
+            }
+
+            val size = icon.layoutParams.height.takeIf { it > 0 }
+                ?.let(IconUtils::getMediaProxySize)
+                ?: return@after
+
             val url = IconUtils.getForGuild(guild, null, animated) + "&size=${size}"
-            MGImages.`setImage$default`(bindingGuild.d, url, size, size, false, null, null, 112, null)
+
+            if (guildIconUrls.put(icon, url) == url) return@after
+
+            MGImages.`setImage$default`(icon, url, size, size, false, null, null, 112, null)
         }
 
         // Support webp emojis by forcing every emoji to be webp
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
@@ -71,6 +71,8 @@ import com.discord.widgets.settings.profile.SettingsUserProfileViewModel
 import com.discord.widgets.settings.profile.WidgetEditUserOrGuildMemberProfile
 import com.discord.widgets.status.WidgetForumPostStatus
 import com.discord.widgets.status.WidgetForumPostStatusViewModel
+import com.discord.widgets.user.profile.UserProfileHeaderView
+import com.discord.widgets.user.profile.UserProfileHeaderViewModel
 import com.discord.widgets.user.usersheet.WidgetUserSheet
 import com.discord.widgets.user.usersheet.WidgetUserSheetViewModel
 import com.linecorp.apng.decoder.Apng
@@ -115,6 +117,7 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
         fixExternalLinks()
         fixClock()
         fixBioHeightLimit()
+        fixProfileBannerColorCrash()
         fixUnreadForumChannels()
         fixMemoryLeak()
     }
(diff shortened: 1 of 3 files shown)


Ushie:
Capping the chatbox to 8000 characters may be excessive, considering that the limit may quickly be reached by pasting a message which the remaining of would be nowhere and that a split messages plugin exists to workaround the message length cap

secp192k1:
> Capping the chatbox to 8000 characters may be excessive, considering that the limit may quickly be reached by pasting a message which the remaining of would be nowhere and that a split messages plugin exists to workaround the message length cap

True, but I chose 8000 for a couple reasons, like getting to that limit in the first place.
Yes using a split messages plugin is ideal honestly, keep in mind that the limit for Nitro is 4000 and Regular is 2000, which the latter is the grand majority of Aliucord users, there would need to be sent 4 separate messages in succession, which can rate limit you - this is something out of scope and should be managed by the split msg plugin, but then again, to keep in mind

secp192k1:
It really does depends on your device as well, of you have bottom of the barrel chinese phone from 2008 and paste 2000 characters you might just freeze.
My take is that the cap shouldn't go over 32000, where 16000 would be already more than enough.

secp192k1:
**6. Fix crash on "NEW! Text chat in voice" popup crash**
When you click on "Try it" right as the popup shows, the activity is `null`, you must wait a while for it to load in the background, then it becomes `com.discord.app.AppActivity$Call@...`

<details>
<summary>image preview of that popup</summary>

text chat in voice crash
</details>

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -514,6 +532,20 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
     }
 
+    private fun fixProfileBannerColorCrash() = tryPatch("Fix profile banner color thread crash") {
+        patcher.before<UserProfileHeaderView>(
+            "updateBannerColor",
+            UserProfileHeaderViewModel.ViewState.Loaded::class.java,
+        ) { param ->
+            if (Looper.myLooper() == Looper.getMainLooper()) return@before
+
+            val view = param.thisObject as UserProfileHeaderView
```
zt64: No need to cast, `before` implicitly casts `this` to `UserProfileHeaderView`

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -514,6 +532,20 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+            val state = param.args[0] as UserProfileHeaderViewModel.ViewState.Loaded
```
zt64: You can use the destructuring functions to cast args[0]

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -648,6 +680,35 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+
+    private val bottomSheetCallbacks = WeakHashMap<BottomSheetBehavior<*>, BottomSheetBehavior.BottomSheetCallback>()
+
+    private fun fixBottomSheetCallbacks() = tryPatch("Fix deprecated BottomSheet callback clearing internal callbacks") {
+        patcher.instead<BottomSheetBehavior<*>>(
+            "setBottomSheetCallback",
+            BottomSheetBehavior.BottomSheetCallback::class.java,
+        ) { param ->
+            @Suppress("UNCHECKED_CAST")
+            val behavior = param.thisObject as BottomSheetBehavior<View>
```
zt64: Same here, use the implicit `this` rather than created another variable. It should work if you change the type param to `BottomSheetBehavior<View>`

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -648,6 +680,35 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+            val callback = param.args[0] as BottomSheetBehavior.BottomSheetCallback?
```
zt64: Use the destructuring function

Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -648,6 +680,35 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
...
+
+            bottomSheetCallbacks.remove(behavior)?.let(behavior::removeBottomSheetCallback)
+            if (callback != null) {
+                behavior.addBottomSheetCallback(callback)
+                bottomSheetCallbacks[behavior] = callback
+            }
```
zt64, line 707: There should be an empty newline before the if statement, and after the closing bracket

zt64 (changes requested):

secp192k1:
~~**7. Fix blurry guild icons**~~

~~The webp guild-icon hook fetches repeatedly, during reorder or fast switching, the~~
~~icon view is unmeasured, so `layoutParams.height` <= `0` downscaled to the~~
~~smallest media url (blurry, low res icons), and the url doesnt change to the full res version~~

Reverted because there is apparently lag. not confident enough to keep it in here

Juby210 (approved):

rushiiMachine:
Is everyone okay with limiting it to 8k characters? it's not that much. Perhaps 15k would be slightly better? It just needs to not crash but being laggy is fine imo

secp192k1:
> @rushiiMachine Is everyone okay with limiting it to 8k characters?

Its not 8k anymore, was changed to 64k (https://github.com/Aliucord/Aliucord/pull/757/changes/6c89c088af5d1b250880b9d8bec9f212496d4159)

Honestly I could just revert this change and keep it unlimited..? Open to suggestions 😄

rushiiMachine:
i think 64k is fine, split messages doesn't allow you to even split that high

PR #765, opened 2026-06-28
fix: removeScrollView now sets NavBar opacity consistently

Importantamigo:
## What's this

This fix targets two main bugs in the Plugin Settings Page both caused by `removeScrollView()` in `settingsPage.java`.

- The first change fixes the bottom padding to not leave extra padding over the Navbar.
- The second change dims the Navbar to be consistent with other settings pages.

<br> 

| Before | After |
|--------|--------|
| Before  | After |

### Aliucord Settings Page for reference

Aliucord Settings


mantikafasi (commented):
# Review for PR #765

I've reviewed the changes in [PR #765](https://github.com/Aliucord/Aliucord/pull/765/changes) which aim to make the SettingsPage's navigation bar transparent/opaque by applying window insets manually.

## 📝 Observations & Potential Issues

The PR introduces the following changes to `com.aliucord.fragments.SettingsPage`:
1. It disables `fitsSystemWindows` on the main view and the linear layout.
2. It manually applies the status bar insets to the `AppBarLayout` and the `LinearLayout`.
3. It sets the bottom padding of the `LinearLayout` to `0`.

### 1. Navigation Bar Overlap (Missing Bottom Inset)
By setting `view.setFitsSystemWindows(false)` and `layout.setFitsSystemWindows(false)`, the entire view is drawn edge-to-edge (behind the status bar and the navigation bar).
However, when manually setting the layout's padding:
```java
layout.setPadding(p, p * 4 + statusBarHeight, p, 0);
```
The bottom padding is set to `0`. Previously, it was `p` (the default padding). Because the view now draws *behind* the navigation bar, setting the bottom padding to `0` will cause content at the bottom of the `SettingsPage` (like buttons or the last items in a list) to be obscured by the system navigation buttons or the gesture navigation pill.

**Recommendation:** You should add the navigation bar height to the bottom padding to prevent content from being hidden behind the navigation bar.

```java
ViewCompat.setOnApplyWindowInsetsListener(view, (v, insets) -> {
    int statusBarHeight = insets.getInsets(WindowInsetsCompat.Type.statusBars()).top;
    int navBarHeight = insets.getInsets(WindowInsetsCompat.Type.navigationBars()).bottom;
    
    appBar.setPadding(0, statusBarHeight, 0, 0);
    layout.setPadding(p, p * 4 + statusBarHeight, p, p + navBarHeight);
    
    return insets;
});
```

### 2. AndroidX API Compatibility
The PR utilizes `WindowInsetsCompat.Type.statusBars()`. This API was introduced in `androidx.core:core:1.5.0`.
Aliucord currently uses `androidx.appcompat:appcompat:1.4.1` (which depends on `core:1.7.0`), so it will compile successfully. However, keep in mind that since `SettingsPage` is a core fragment, it heavily depends on the AndroidX Core version shipped with the specific Discord versions Aliucord supports at runtime. As long as the targeted Discord versions bundle `androidx.core:core:1.5.0` or higher, this will not result in a `NoSuchMethodError`.

### 3. JavaDoc Formatting
The PR cleans up the formatting for the `@param text` JavaDoc in `addHeader`, which is a nice minor improvement.

## Conclusion

The approach to manually handle window insets works for making the status/navigation bar transparent, but the **bottom padding must be adjusted** to include `WindowInsetsCompat.Type.navigationBars().bottom` so that content isn't drawn directly underneath the system navigation bar.

Importantamigo:
- Added bottom padding to the recycler view to stop cards from being drawn under NavBar.
- set `setClipToPadding` to `false` to let recyclerView draw over padding

.github/workflows/fetch-contributors.yml
rushiiMachine, line 1: seems like you accidentally added this to your pr

rushiiMachine (changes requested):

Importantamigo:
aww fuck me

rushiiMachine (approved):

PR #745, opened 2026-06-07
feat(smali): Add all missing fields

mantikafasi:
## Summary

- Add missing Discord message object fields
- Add smali field patches for:
  - `Message`
  - model `Message`
  - `MessageEmbed`
  - embed media objects
  - `MessageReaction`
  - `MessageAttachment`
- Add Injector DTOs for included/nested objects:
  - reaction count details
  - content scan metadata
  - purchase notification / guild product purchase
  - gift info / message soundboard sound
  - potion
  - shared client theme
  - soundboard sound
- Preserve newly added public fields when `com.discord.models.message.Message.copy(...)` is used

## Validation

- `./gradlew :patches:testPatches` passes

PR #769, opened 2026-07-06
feat: add missing fields for guild feature and onboarding

Canny1913:
Adds missing fields for Onboarding functionality and adds rest of missing guild features.

### DIFF
--- a/patches/src/com/discord/api/guild/GuildFeature.patch
+++ b/patches/src/com/discord/api/guild/GuildFeature.patch
@@ -1,0 +1,55 @@
+--- smali_original/com/discord/api/guild/GuildFeature.smali
++++ smali/com/discord/api/guild/GuildFeature.smali
+@@ -130,12 +130,16 @@
+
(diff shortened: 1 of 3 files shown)


Ushie:
you say work bu yes?

Canny1913:
> you say work bu yes?

i say yes it does (i think)

rushiiMachine:
Doesn't this conflict with #745 ? @mantikafasi which one should be merged first

mantikafasi:
> Doesn't this conflict with #745 ? @mantikafasi which one should be merged first

well mine came wayyy before so imo mine it is. also I believe they are patching different classes. mine did not patch onboarding

secp192k1:
> Doesn't this conflict with #745 ? @mantikafasi which one should be merged first

There is no collision between that PR and this one, also, this now adds all missing Guild Features instead of only adding the onboarding ones

rushiiMachine (approved):

PR #771, opened 2026-07-08
fix(patches): switch GIF provider to Klipy

insertusernamed:
Tenor API shut down June 30, 2026. Discord already migrated their servers to accept provider=klipy. This patch replaces the 5 hardcoded `"tenor"` strings in `StoreGifPicker` with `"klipy"` so GIF search keeps working.

Verified with `testPatches` - all patches apply cleanly. Tested on device via custom patches import.

### DIFF
--- a/patches/src/com/discord/stores/StoreGifPicker.patch
+++ b/patches/src/com/discord/stores/StoreGifPicker.patch
@@ -1,0 +1,46 @@
+--- smali_original/com/discord/stores/StoreGifPicker.smali
++++ smali/com/discord/stores/StoreGifPicker.smali
+@@ -793,7 +793,7 @@
+
+     move-result-object v2
+
+-    const-string/jumbo v3, "tenor"
++    const-string/jumbo v3, "klipy"
+
+     const/4 v4, 0x5
+
+@@ -855,7 +855,7 @@
+
+     move-result-object v3
+
+-    const-string/jumbo v2, "tenor"
++    const-string/jumbo v2, "klipy"
+
+     const-string/jumbo v4, "tinygif"
+
+@@ -934,7 +934,7 @@
+
+     move-result-object v1
+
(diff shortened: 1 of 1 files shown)


secp192k1:
Insane implementation method but fear not, discord's code is worse:

image
## none of these variables are used, guh

Juby210 (approved):

zt64 (approved):

Ushie (approved):

mantikafasi (changes requested):
I dont think smali patching is required here, simple patch to RestAPI would be sufficent

insertusernamed:
5 string replacements in StoreGifPicker vs 5 methods with conditional logic in RestAPI. Both require patching the same number of sites but the StoreGifPicker patch is just literal swaps with no branching. Keeps the diff minimal and avoids introducing new code paths.

mantikafasi:
only issue is you need a full reinstall to make this work, whereas if it was restapi patch with 2 patches you could achieve same thing

secp192k1:
> only issue is you need a full reinstall to make this work, whereas if it was restapi patch with 2 patches you could achieve same thing

(As talked in dms) Orrrr move this to a core fix, which defeats the need to repatch/reinstall aliucord entirely

Ushie:
The requirement to update the Aliucord app through Aliucord Manager is besides the point, as version bump isn't going to be part of this PR and other smali changes can be part of the same incoming release (i.e. the other PR that adds essential things like the message attachment flags for the new spoilers and new message types)

So this should be fine

mantikafasi:
BETTER MERGE https://github.com/Aliucord/Aliucord/pull/745 BEFORE THIS THEN

secp192k1:
> BETTER MERGE #745 BEFORE THIS THEN

larp

wingio (approved):

rushiiMachine (approved):

PR #782, opened 2026-08-12
fix(CoreFixes): Nitro emojis not working in user status

secp192k1:


Aliucord/src/main/java/com/aliucord/coreplugins/CoreFixes.kt
```diff
@@ -747,6 +750,32 @@ internal class CoreFixes : CorePlugin(Manifest("CoreFixes")) {
         }
     }
 
+    private fun fixCustomStatusEmoji() = tryPatch("Fix nitro emojis in user status") {
+        var pendingEmoji: ModelEmojiCustom? = null
```
rushiiMachine, line 754: This is never released?

PR #783, opened 2026-08-20
feat(rn): handle RN spoilers

Archimedes9500:
Discord adding some bullshit, spoilered attachments now might be marked with a flag (`1<<3`) and not have `SPOILER_` in the filename at all

### DIFF
--- a/Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
+++ b/Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
@@ -294,7 +294,7 @@ val MessageAttachment.flags: Int by accessField();
 fun patchSpoilers(){
     Patcher.addPatch(MessageAttachment::class.java.getDeclaredMethod("h"/*"isSpoiler"*/), Hook {
         val isSpoiler = it.result as Boolean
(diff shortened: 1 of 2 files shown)


Aliucord/src/main/java/com/aliucord/coreplugins/rn/Patches.kt
```diff
@@ -287,3 +289,14 @@ fun patchAuditLog() {
...
+val MessageAttachment.flags: Int by accessField();
+fun patchSpoilers(){
+    Patcher.addPatch(MessageAttachment::class.java.getDeclaredMethod("h"/*"isSpoiler"*/), Hook {
+        val isSpoiler = it.result as Boolean
+        if (isSpoiler || it.thisObject == null) return@Hook
+        with (it.thisObject as MessageAttachment) {
+            it.result = (0 != (flags and 8/*MessageAttachment.SPOILER*/))
+        }
+    })
+}
```
rushiiMachine, line 302: This code is actually horrendous

Archimedes9500: How else you want to check for flag 🥀

PR #785, opened 2026-09-01
fix: set new max upload size

### LINKED ISSUE #786: Change maximum upload size to 20MB

_No response_

### What’s the issue?

Discord has increased the base upload size to 20MB, yet the app stillimits it to 10

I'm able to upload files over 10MB

### How can we reproduce it?

Try to upload a file over 10MB

### Android Version

17

### Device & OS Info

GrapheneOS on Pixel 8 Pro

### Crash Log (if applicable)

_No response_

### Before Submitting

Importantamigo:
Braindead change that makes the new upload limits work, tested 
Screenshot_20260901-134418_Aliucord


Aliucord/src/main/java/com/aliucord/coreplugins/UploadSize.kt
secp192k1, line 106: While you're at it, maybe fix this too :blobcatcozy:

Ushie: Idk why that even exists when the argument is set by us to begin with

Importantamigo: oh shii mb I'll fix

Importantamigo: > While you're at it, maybe fix this too :blobcatcozy:

it doesn't seem to break anything, images bigger than 8mb still work fine

cillynder: This doesn't really matter; it's used for the error string that ends up being replaced anyway

Questionable implementation but it could be refactored in a separate PR

cillynder (approved):
lgtm

